PR #1, opened 2019-10-17
C2LC-16: ClapDetector

chosww:


src/ClapDetector.js
```diff
@@ -0,0 +1,168 @@
...
+                            this.props.onDetect('left');
+                        } else if (sim2 >= sim1 && sim2 >= sim3 && sim2 >= 0.90) {
+                            console.log(sim2);
+                            console.log('right');
+                            this.props.onDetect('right');
+                        } else if (sim3 >= sim1 && sim3 >= sim2 && sim3 >= 0.90) {
+                            console.log(sim3);
+                            console.log('forward');
+                            this.props.onDetect('forward');
+                        }
```
simonbates: We can store the analysers anonymously and loop through to find the one with the highest similarity.

src/ClapDetector.js
```diff
@@ -0,0 +1,168 @@
...
+
+                Promise.all([
+                    this.addAudioStreamNode(this.state.audioInputDeviceIds[0], context),
+                    this.addAudioStreamNode(this.state.audioInputDeviceIds[1], context),
+                    this.addAudioStreamNode(this.state.audioInputDeviceIds[3], context)
+                ]).then((analysers) => {
+                    const defaultInputAnalyser = analysers[0];
+                    const externalInputAnalyser = analysers[1];
+                    const wirelessInputAnalyser = analysers[2];
+                    const scriptProcessorAnalysisNode = context.createScriptProcessor(2048,2,1);
```
simonbates: According the MDN, this function is deprecated:

https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/createScriptProcessor

simonbates: Why 2 input channels?

src/ClapDetector.js
simonbates: Could we check for other qualities here such as which one is loudest?

src/ClapDetector.js
```diff
@@ -0,0 +1,168 @@
...
+                            console.log('forward');
+                            this.props.onDetect('forward');
+                        }
+                    }
+                });
+            }
+        }
+    }
+
+    addAudioStreamNode = async (deviceId: string, audioContext: any): any=> {
```
simonbates: I think `deviceId` isn't used and we always use `this.state.audioInputDeviceIds[0]`

src/ClapDetector.js
```diff
@@ -0,0 +1,168 @@
...
+                    this.addAudioStreamNode(this.state.audioInputDeviceIds[1], context),
+                    this.addAudioStreamNode(this.state.audioInputDeviceIds[3], context)
+                ]).then((analysers) => {
+                    const defaultInputAnalyser = analysers[0];
+                    const externalInputAnalyser = analysers[1];
+                    const wirelessInputAnalyser = analysers[2];
+                    const scriptProcessorAnalysisNode = context.createScriptProcessor(2048,2,1);
+                    const audioGain = context.createGain();
+                    scriptProcessorAnalysisNode.connect(audioGain);
+                    audioGain.connect(context.destination);
```
simonbates: Reading through the audio code here, I see these connections:

- (`scriptProcessorAnalysisNode`)-(`audioGain`)-(`context.destination`) [above]
- (`audioInputStream`)-(`audioAnalyser`) [in `addAudioStreamNode()` below]

Is there a connection between these 2 groups of nodes?

src/ClapDetector.js
```diff
@@ -0,0 +1,168 @@
...
+            analyser : audioAnalyser,
+            freqBuffer : arrayFreqDomain,
+            timeBuffer : arrayTimeDomain
+        });
+    }
+
+    getClapSimilarity(analyserNode: any, numRows: number, deviceId: number): number {
+        let timeDomainBuffer = analyserNode.analyser.getByteTimeDomainData(analyserNode.timeBuffer);
+        let freqDomainBuffer = analyserNode.analyser.getByteFrequencyData(analyserNode.freqBuffer);
+        let currentTimeValue = (analyserNode.timeBuffer[0] / 128) - 1.0;
```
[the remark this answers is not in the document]
chosww: getByteTimeDomainData (and the newer getFloatTimeDomainData) return an array of the size you requested - its frequencyBinCount, which is calculated as half of the requested fftSize. That array is, of course, at the current sampleRate exposed on the AudioContext, so if it's the default 2048 fftSize, frequencyBinCount will be 1024, and if your device is running at 44.1kHz, that will equate to around 23ms of data.

The byte values do range between 0-255, and yes, that maps to -1 to +1, so 128 is zero. (It's not volts, but full-range unitless values.)

If you use getFloatFrequencyData, the values returned are in dB; if you use the Byte version, the values are mapped based on minDecibels/maxDecibels (see the minDecibels/maxDecibels description).

--https://stackoverflow.com/questions/24083349/understanding-getbytetimedomaindata-and-getbytefrequencydata-in-web-audio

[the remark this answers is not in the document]
chosww: Since analyserNode.timeBuffer array contains values in range between 0 to 255 or -1 to +1 accordingly, by dividing the value from the buffer by 128 and -1 will return actual value; for example, 128 should be 0, so (128 / 128) - 1.0 = 0.

src/ClapDetector.js
```diff
@@ -0,0 +1,168 @@
...
+        let similarity = 0;
+        if (Math.abs(currentTimeValue) !== 0.0078125
+        && currentTimeValue !== 0 && Math.abs(currentTimeValue) !== 0.015625
+        && Math.abs(currentTimeValue) !== 0.0234375) {
```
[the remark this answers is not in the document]
chosww: These numbers are returned when the timeBuffer values are meaninglessly small; currentTimeValue will be equal to 0.0078125, when the timeBuffer value is equal to 129, 0.015625 when the value is equal to 130, and 0.0234375 when value is equal to 131 (note that the buffer value 128 means 0). These values were observed when there's no sound inputs were detected, so I filtered those results out.

src/ClapDetector.js
chosww: Since they are created from the same AudioContext, they seem to have implicit connection.

src/App.js
```diff
@@ -155,6 +177,11 @@ export default class App extends React.Component<{}, AppState> {
         );
     }
 
+    componentDidMount() {
+        this.soundAnalyser.getMicrophones();
+        this.soundAnalyser.configureDetection({"cb3b7c5cefb4536a35cdb8ab6975a00705300108fe461fc31965de349f3722d7" : this.handleClap});
```
simonbates: We aren't using the discovered microphones. I'm thinking "cb3b7c5cefb4536a35cdb8ab6975a00705300108fe461fc31965de349f3722d7" is an id from your laptop?

src/C2lcMath.js
```diff
@@ -4,8 +4,27 @@ function degrees2radians(degrees: number): number {
     return degrees * Math.PI / 180;
 }
 
+function getNormalizedCorrelation(x: Array<number>, y: Array<number>): number {
+    try {
+        if (x.length !== y.length) throw 'Length of x and y should be equal';
+    } catch(err) {
```
simonbates: We should be throwing the exception upwards, rather than catching it here.

src/C2lcMath.test.js
```diff
@@ -27,3 +27,10 @@ test('C2lcMath.wrap', () => {
     expect(C2lcMath.wrap(-20, -10, -30)).toBe(-20);
     expect(C2lcMath.wrap(-20, -10, -33)).toBe(-13);
 });
+
+test('Calculating normalized correlation', () => {
+    // javascript rounds-up to 16 decimal degits 
+    expect(C2lcMath.getNormalizedCorrelation([0,2,3],[0,2])).toBe(0);
```
simonbates: We expect an exception to be thrown here, rather than a return value of 0.

src/SoundAnalyser.js
```diff
@@ -0,0 +1,90 @@
...
+        const arrayTimeDomain = new Uint8Array(bufferLength);
+        return ({
+            deviceId,
+            analyser : audioAnalyser,
+            freqBuffer : arrayFreqDomain,
+            timeBuffer : arrayTimeDomain
+        });
+    }
+
+    start() {}
```
simonbates: Needs an implementation

src/SoundAnalyser.js
```diff
@@ -0,0 +1,90 @@
...
+    
+    stop() {}
```
simonbates: Needs an implementation

src/SoundAnalyser.js
```diff
@@ -0,0 +1,90 @@
...
+    constructor(webAudioContextInstance: any, mediaDevicesInstance: any, soundSamples:  { clap : Array<number> }) {
+        this.webAudioContextInstance = webAudioContextInstance;
+        this.mediaDevicesInstance = mediaDevicesInstance;
+
+        this.deviceList = [];
+        this.context = new webAudioContextInstance();
+        this.soundSamples = soundSamples;
+    }
+
+    getMicrophones = () => {
```
simonbates: This method records the microphones in an array `deviceList` but that array is never used. I think we should instead return a Promise with the list of microphone ids.

simonbates: Thinking about the name here. `getAudioInputDevices` might be a better name.

src/C2lcMath.test.js
```diff
@@ -27,3 +27,10 @@ test('C2lcMath.wrap', () => {
     expect(C2lcMath.wrap(-20, -10, -30)).toBe(-20);
     expect(C2lcMath.wrap(-20, -10, -33)).toBe(-13);
 });
+
+test('Calculating normalized correlation', () => {
+    // javascript rounds-up to 16 decimal degits
+    expect(() => {C2lcMath.getNormalizedCorrelation([0,2,3],[0,2])}).toThrow();
```
simonbates: I think we don't need the anonymous function here. I think just:

```
expect(C2lcMath.getNormalizedCorrelation([0,2,3],[0,2])).toThrow();
```

simonbates: And we can also check the error message:

```
expect(C2lcMath.getNormalizedCorrelation([0,2,3],[0,2])).toThrow('Length of x and y should be equal');
```

PR #2, opened 2019-10-30
C2LC-19 code review

chosww:


src/App.js
```diff
@@ -31,7 +36,9 @@ type AppState = {
     programVer: number,
     settings: AppSettings,
     dashConnectionStatus: DeviceConnectionStatus,
-    speechRecognitionOn: boolean
+    mode: string,
```
simonbates: Is this the state value that is controlling which editor is visible? Let's name it something more descriptive like `editorMode`. Let's also move it inside `settings` and add a new type to constrain the allowed values:

```
type EditorMode = 'block' | 'text';

type AppSettings = {
    dashSupport: boolean,
    editorMode: EditorMode,
    language: string
}
```

PR #4, opened 2019-11-28
C2LC-38: Setup CD pipeline for dev environment

chosww:


simonbates:
Please update `.firebaserc` and `firebase.json` to use 4-space indents, to be consistent with our other JSON files (`package.json`).

.github/workflows/ci.yaml
```diff
@@ -34,7 +34,28 @@ jobs:
...
+          path: build
       - name: Flow
         run: npx flow
       - name: Test
         run: npm test
+  deploy:
+    name: Deploy
+    needs: build
+    runs-on: ubuntu-18.04
+    steps:
```
simonbates: Please add a check here, above `steps`, to only run on the project repo, the `dev-0.1` branch, and only on `push` events:

```
if: (github.repository == 'codelearncreate/c2lc-coding-environment') && (github.ref == 'refs/heads/dev-0.1') && (github.event_name == 'push')
```

firebase.json
```diff
@@ -0,0 +1,16 @@
...
+    "ignore": [
+      "firebase.json",
+      "**/.*",
+      "**/node_modules/**"
+    ],
+    "rewrites": [
+      {
+        "source": "**",
+        "destination": "/index.html"
+      }
```
simonbates: I think for now I'd prefer to not do rewriting here. And let any requests to missing resources just 404. If we add retrieving data from the URL, we can think about it then.

PR #5, opened 2019-11-29
C2LC-37: Remove components not used in 0.1 from App.js

chosww:
Removed unused components and states for 0.1 release. 

Some update to a handler 'handleClickRun()' to check if dash is connected first:

`
handleClickRun = () => {
    if (this.state.dashConnectionStatus === 'connected') {
        this.interpreter.run(this.state.program);
    }
};
`

Without checking the connection, errors will occur, since the interpreter doesn't know any commands until dash is connected from componentDidUpdate() in App.js

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -27,8 +27,7 @@ type AppContext = {
 };
 
 type AppSettings = {
-    language: string,
-    dashSupport: boolean
+    language: string
 }
 
 type AppState = {
@@ -62,8 +61,7 @@ export default class App extends React.Component<{}, AppState> {
                 'left'
             ],
             settings: {
-                language: 'en',
-                dashSupport: this.appContext.bluetoothApiIsAvailable
+                language: 'en'
             },
             dashConnectionStatus: 'notConnected',
             selectedAction: null
@@ -72,7 +70,23 @@ export default class App extends React.Component<{}, AppState> {
         this.interpreter = new Interpreter();
 
         this.dashDriver = new DashDriver();
+
+        this.interpreter.addCommandHandler(
+            'none',
+            'noneHandler',
+            () => {
+                return Promise.resolve();
+            }
+        );
     }
+
+    setStateSettings(settings: $Shape<AppSettings>) {
+        this.setState((state) => {
+            return {
+                settings: Object.assign({}, state.settings, settings)
+            }
+        });
+    };
 
     getSelectedCommandName() {
         if (this.state.selectedAction !== null
@@ -81,7 +95,7 @@ export default class App extends React.Component<{}, AppState> {
         } else {
             return null;
         }
-    }
+    };
 
     handleChangeProgram = (program: Program) => {
         this.setState({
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -74,87 +62,16 @@ export default class App extends React.Component<{}, AppState> {
...
-                }
-            }
-        );
-        this.interpreter.addCommandHandler(
-            'none',
-            'noneHandler',
-            () => {
-                return Promise.resolve();
-            }
-        );
```
simonbates: Let's keep the 'none' handler

src/App.js
```diff
@@ -166,26 +83,16 @@ export default class App extends React.Component<{}, AppState> {
...
     };
 
     handleClickRun = () => {
-        if (this.turtleGraphicsRef.current !== null) {
-            this.turtleGraphicsRef.current.clear();
-        }
-        if (this.turtleGraphicsRef.current !== null) {
-            this.turtleGraphicsRef.current.home();
+        if (this.state.dashConnectionStatus === 'connected') {
+            this.interpreter.run(this.state.program);
```
simonbates: As discussed on IRC, let's instead disable the "Run" button when dashConnectionStatus !== 'connected'

src/App.js
```diff
@@ -74,87 +62,16 @@ export default class App extends React.Component<{}, AppState> {
...
-                this.handleSpeechCommand);
-        }
-    }
-
-    setStateSettings(settings: $Shape<AppSettings>) {
-        this.setState((state) => {
-            return {
-                settings: Object.assign({}, state.settings, settings)
-            }
-        });
```
simonbates: Let's keep setStateSettings (even though it's not currently used)

src/App.js
```diff
@@ -74,87 +62,16 @@ export default class App extends React.Component<{}, AppState> {
                 'left'
             ],
             settings: {
-                dashSupport: this.appContext.bluetoothApiIsAvailable,
-                editorMode: 'block',
-                language: 'en'
+                language: 'en',
+                dashSupport: this.appContext.bluetoothApiIsAvailable
```
simonbates: Let's remove the `dashSupport` setting and always show the "Connect to Dash" control.

https://issues.fluidproject.org/browse/C2LC-91 will make the control disabled when Bluetooth isn't available

src/App.js
```diff
@@ -28,38 +23,31 @@ import './App.css';
...
-    speechRecognitionApiIsAvailable: boolean
+    bluetoothApiIsAvailable: boolean
 };
 
 type AppSettings = {
-    dashSupport: boolean,
-    editorMode: EditorMode,
-    language: string
+    language: string,
+    dashSupport: boolean
```
simonbates: Let's remove the `dashSupport` setting and always show the "Connect to Dash" control.

https://issues.fluidproject.org/browse/C2LC-91 will make the control disabled when Bluetooth isn't available

PR #7, opened 2019-12-02
C2LC-91: Disable 'Connect to Dash' button when bluetooth API is not supported

chosww:
Tested on Safari to check if the button is disabled.


simonbates:
Please merge the latest dev-0.1 into C2LC-91 and resolve the conflicts

src/DeviceConnectControl.js
```diff
@@ -11,6 +11,7 @@ import './DeviceConnectControl.css'
 type DeviceConnectControlProps = {
     intl: any,
     children: React.Element<any>, // Button contents
+    bluetoothApiIsAvailable: boolean,
```
simonbates: Could we instead rename this property to "disabled". I'm thinking that a more general disabled property is useful and we don't need to specify the reason why it's disabled in the API. The control doesn't vary based on the reason why it's disabled, only that it's disabled.

PR #9, opened 2019-12-03
C2LC-86: Update styling and layout of the command palette

chosww:
- Remove tab style UI, just display commands on a Col component. 
- Update layout of components, according to 0.1 design. 
- Need to update styling when everything merges together.
- Remove localizeProperties function as well as injectIntl module since we are not injecting any text now.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -8,7 +8,14 @@
 }
 
 .App__command-palette {
+    background-color: #F1F1F1;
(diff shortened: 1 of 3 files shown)


chosww:
Seems like I worked on top of C2LC-91 instead of C2LC-37

chosww:
Merged to C2LC-87

PR #13, opened 2019-12-04
C2LC-85: Show Dash connection error modal

chosww:


src/App.js
```diff
@@ -33,6 +34,7 @@ type AppState = {
     program: Program,
     settings: AppSettings,
     dashConnectionStatus: DeviceConnectionStatus,
+    showError: boolean,
```
simonbates: Please rename to `showDashConnectionError`

src/ConnectionErrorModal.js
```diff
@@ -0,0 +1,64 @@
...
+import './ConnectionErrorModal.css';
+
+type ConnectionErrorModalProps = {
+    intl: any,
+    show: boolean,
+    onCancel: () => void,
+    onRetry: () => void
+};
+
+class ConnectionErrorModal extends React.Component<ConnectionErrorModalProps, {}> {
```
simonbates: The messaging on this component is Dash-specific, so let's rename the component to `DashConnectionErrorModal`.

PR #17, opened 2019-12-09
C2LC-91: Add test cases for DeviceConnectControl

chosww:


src/DeviceConnectControl.test.js
```diff
@@ -0,0 +1,94 @@
+// @flow
+
+import React from 'react';
+import Adapter from 'enzyme-adapter-react-16';
+import { shallow, mount, configure } from 'enzyme';
```
simonbates: We don't need `mount`, I think

src/DeviceConnectControl.test.js
```diff
@@ -0,0 +1,94 @@
...
+import { Col, Button } from 'react-bootstrap';
```
simonbates: I think we don't need either of these `react-bootstrap` imports

src/DeviceConnectControl.test.js
```diff
@@ -0,0 +1,94 @@
...
+    );
+    expect(hasDisabledClass(wrapper)).toBe(true);
+    expect(buttonIsDisabled(wrapper)).toBe(true);
+});
+
+test('Checking icon and aria-label for notConnected status', () => {
+    const wrapper = shallow(
+        <DeviceConnectControl.WrappedComponent
+            intl={intl}
+            disabled={true}
```
simonbates: Small detail, but let's do this test with disabled={false}

src/DeviceConnectControl.test.js
```diff
@@ -0,0 +1,94 @@
...
+    expect(getStatusIconContainer(wrapper).props['aria-label']).toBe(intl.messages['DeviceConnectControl.notConnected']);
+    expect(getStatusIconContainer(wrapper).props.role).toBe('img');
+    expect(getStatusIconContainer(wrapper).props.children.type.render.name).toBe('SvgStatusNotConnected')
+});
+
+test('Checking icon and aria-label for connected status', () => {
+    const wrapper = shallow(
+        <DeviceConnectControl.WrappedComponent
+            intl={intl}
+            disabled={true}
```
simonbates: Small detail, but let's do this test with disabled={false}

src/DeviceConnectControl.test.js
```diff
@@ -0,0 +1,94 @@
...
+    expect(getStatusIconContainer(wrapper).props['aria-label']).toBe(intl.messages['DeviceConnectControl.connected']);
+    expect(getStatusIconContainer(wrapper).props.role).toBe('img');
+    expect(getStatusIconContainer(wrapper).props.children.type.render.name).toBe('SvgStatusConnected')
+});
+
+test('Checking icon and aria-label for connecting status', () => {
+    const wrapper = shallow(
+        <DeviceConnectControl.WrappedComponent
+            intl={intl}
+            disabled={true}
```
simonbates: Small detail, but let's do this test with disabled={false}

PR #24, opened 2019-12-12
C2LC-100: Fix behaviour problem during run

chosww:
This request is a fix to behaviour problems those can occur while program is running. I used newly introduced state from C2LC-84, which still have an open pull request. It would probably make sense to either merge C2LC-84 first, or close the pull request for C2LC-84 and review this one. -- I updated the pull request for C2LC-84 with test cases for auto scrolling.


src/Interpreter.test.js
```diff
@@ -141,6 +142,58 @@ test('Do a command with a program', (done) => {
 });
 
 test('Doing an unknown command rejects with Error', () => {
-    const interpreter = new Interpreter();
+    const interpreter = new Interpreter(()=>{});
     return expect(interpreter.doCommand('unknown-command')).rejects.toThrow('Unknown command: unknown-command');
 });
+
+test('Use the callback function on continueRun', () => {
```
simonbates: Tests calling `run()` need to be async as `run()` is async: add a parameter to the test function called `done` and call `done()` at the end of the assertions to indicate to the test framework that the async test has completed.

src/Interpreter.test.js
```diff
@@ -141,6 +142,58 @@ test('Do a command with a program', (done) => {
...
+    interpreter.run(['step1', 'step2']).then(() => {
+        expect(mockStateChangeHandler.mock.calls.length).toBe(3);
+        expect(mockStateChangeHandler.mock.calls[0][0]).toStrictEqual({'isRunning': true, 'activeStep': 0});
+        expect(mockStateChangeHandler.mock.calls[1][0]).toStrictEqual({'isRunning': true, 'activeStep': 1});
+        expect(mockStateChangeHandler.mock.calls[2][0]).toStrictEqual({'isRunning': false, 'activeStep': null});
+    });
+});
+
+
+test('Do not continue through program if stop is called', () => {
```
simonbates: Tests calling `run()` need to be async as `run()` is async: add a parameter to the test function called `done` and call `done()` at the end of the assertions to indicate to the test framework that the async test has completed.

src/Interpreter.test.js
simonbates: Please make 3 variants of this test:

- An empty program: "onRunningStateChange is called on run() empty program"
- A program with one step: "onRunningStateChange is called on run() program with one step"
- A program with two steps "onRunningStateChange is called on run() program with two steps"

src/Interpreter.js
```diff
@@ -39,22 +42,30 @@ export default class Interpreter {
...
+                this.onRunningStateChange({isRunning: this.isRunning, activeStep: null});
+                resolve();
             } else {
+                this.onRunningStateChange({isRunning: this.isRunning, activeStep: this.programCounter});
                 this.step().then(() => {
-                    this.continueRun();
+                    this.continueRun(resolve, reject);
                 });
+                // TODO: reject run promise if step promise rejected
             }
```
simonbates: Address the TODO comment by calling `reject` when the `step()` Promise is rejected:

````
} else {
    this.onRunningStateChange({isRunning: this.isRunning, activeStep: this.programCounter});
    this.step().then(() => {
        this.continueRun(resolve, reject);
    }, () => {
        // Reject the run Promise when the step Promise is rejected
        reject();
    });
}
````

And add a test case to `Interpreter.test.js` to verify that the run Promise is rejected.

For an example of asserting a Promise rejection, see: https://github.com/codelearncreate/c2lc-coding-environment/blob/6d770ca4e32482e6d657d1b1497c941fdf0bdf19/src/Interpreter.test.js#L95-L100

File: src/EditorContainer.js [hunk 4]
```diff
@@ -28,6 +28,7 @@ export default class EditorContainer extends React.Component<EditorContainerProp
                         onChange={this.props.onChange} />
                  ) : (
                     <ProgramBlockEditor
+                        activeProgramStepNum={-1}
```
simonbates: Set to `null` rather than -1

src/ProgramBlockEditor.js
```diff
@@ -25,6 +26,9 @@ type ProgramBlockEditorProps = {
 };
 
 class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, {}> {
+    commandBlock: Button
+    programSequenceWindow: Col
```
simonbates: It looks like `programSequenceWindow` isn't used

src/ProgramBlockEditor.js
```diff
@@ -82,13 +86,22 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, {}> {
...
+            'command-block'
+        ];
+        if (active) {
+            classNames.push('ProgramBlockEditor__program-block--active');
+        }
         switch(command) {
             case 'forward':
                 return (
                     <Button
+                        ref={active ? r => (this.commandBlock = r) : null}
```
simonbates: Please move the logic from this anonymous function to a method on ProgramBlockEditor: `setActiveCommandBlockRef` and then the JSX becomes:

```
ref={this.setActiveCommandBlockRef}
```

For an example, see: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

src/ProgramBlockEditor.js
```diff
@@ -189,6 +205,7 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, {}> {
                     </Col>
                     <div className='ProgramBlockEditor__editor-actions'>
                         <Button
+                            disabled={this.props.activeProgramStepNum !== null}
```
simonbates: I'd like to make this logic more explicit: let's add a new prop to `ProgramBlockEditor` called `editingDisabled` and put the logic for disabling (`interpreterIsRunning === true`) in `App.js`

src/ProgramBlockEditor.js
```diff
@@ -25,6 +26,9 @@ type ProgramBlockEditorProps = {
 };
 
 class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, {}> {
+    commandBlock: Button
```
simonbates: Please add a constructor to initialize commandBlock:

    constructor(props: ProgramBlockEditorProps) {
        super(props);
        this.commandBlock = null;
    }

src/ProgramBlockEditor.test.js
```diff
@@ -233,3 +240,77 @@ test('blocks', () => {
...
+    expect(mockScrollInto.mock.calls[0][0]).toStrictEqual({ behavior: 'smooth', block: 'start', inline: 'start' });
+
+    wrapper.setProps({ activeProgramStepNum: 2});
+
+    expect(mockScrollInto.mock.calls.length).toBe(2);
+    expect(mockScrollInto.mock.calls[1][0]).toStrictEqual({ behavior: 'smooth', block: 'start', inline: 'start'});
+
+});
+
+test('When the program is running (if activeProgramStepNum prop has a value), editor action buttons should be disabled', () => {
```
simonbates: The logic to disable no longer depends on activeProgramStepNum -- update the test title:

'The editor action buttons disabled states are set according to the editingDisabled property'

src/ProgramBlockEditor.test.js
```diff
@@ -233,3 +240,77 @@ test('blocks', () => {
...
+            wrappingComponent: IntlProvider,
+            wrappingComponentProps: {
+                locale: 'en',
+                defaultLocale: 'en',
+                messages: messages.en
+            }
+        }
+    );
+
+    // activeProgramStepNum is null
```
simonbates: // editingDisabled is false

src/ProgramBlockEditor.test.js
```diff
@@ -233,3 +240,77 @@ test('blocks', () => {
...
+    expect(getEditorActionButtons(wrapper).get(0).props.disabled).toBe(false);
+    expect(getEditorActionButtons(wrapper).get(1).props.disabled).toBe(false);
+
+    wrapper.setProps({editingDisabled: true});
+
+    // activeProgramStepNum is 0
```
simonbates: Remove this comment

PR #28, opened 2019-12-20
C2LC-103: Show the Run button as inverted when a program is running

chosww:


chosww:
@simonbates Current implementation uses editingDisabled property to decide whether program is running or not, which will change state of run button to pressed. We could pass interpreterIsRunning as a prop to ProgramBlockEditor if we want the logic to be more clear.

simonbates:
Yes, I think passing an explicit `interpreterIsRunning` is clearer and will be preferable to maintain. Thanks.

PR #37, opened 2020-01-08
C2LC-113: Adding and deleting are possible while program is running

chosww:


simonbates:
This implementation will cause the user to lose their `selectedAction` upon running the program.

I'm thinking that a better user experience would be to instead leave the `selectedAction` unchanged and extend the application of `editingDisabled` to disable everything that could cause a program edit, in particular the program blocks.

This will mean changing the program blocks from `Button` into `AriaDisablingButton`.

src/ProgramBlockEditor.scss
```diff
@@ -131,6 +131,10 @@ button.ProgramBlockEditor__program-block {
     margin-right: 1.4rem;
 }
 
+button.ProgramBlockEditor__program-block--disabled {
+    fill: $c2lc-disabled-color;
+}
+
```
simonbates: We don't need this new rule as we want disabled program blocks to look unchanged when disabled.

simonbates: We don't need this new rule as we want disabled program blocks to look unchanged when disabled.

src/ProgramBlockEditor.js
```diff
@@ -150,78 +150,89 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
...
+                        data-stepnumber={programStepNumber}
                         aria-label={
                             this.addIsSelected() ?
                             `${this.props.intl.formatMessage({id:'ProgramBlockEditor.commandForward'}, {index: programStepNumber + 1})}. ${this.props.intl.formatMessage({id:'ProgramBlockEditor.commandOnAdd'})}` :
                             this.deleteIsSelected() ?
                             `${this.props.intl.formatMessage({id:'ProgramBlockEditor.commandForward'}, {index: programStepNumber + 1})}. ${this.props.intl.formatMessage({id:'ProgramBlockEditor.commandOnDelete'})}` :
                             this.props.intl.formatMessage({id:'ProgramBlockEditor.commandForward'}, {index: programStepNumber + 1})
                         }
-                        onClick={this.handleClickStep}>
+                        disabledClassName='ProgramBlockEditor__program-block--disabled'
```
simonbates: Please make `disabledClassName` optional on `AriaDisablingButton` and remove the attribute from the program blocks. We want to set `aira-disabled` on the program blocks, but we don't need any presentation change.

PR #39, opened 2020-01-09
C2LC-117: Scroll to reveal the end of the program for adding blocks

chosww:
Autoscrolling to the last 'none' block was done by always setting the reference to 'none' blocks in makeProgramBlock function, since 'none' block always needs to be focused whether when it's newly created or activeStep is a 'none' block. I couldn't add test cases for auto scrolling, because the last none block is for UI, so can't access it programmatically with an index. 

For the scrollbar styling, the scrollbar thumb's width depends on what's inside of the scrollable window, so visibly it is off from the new design.


simonbates:
@chosww I've made a minor update to C2LC-117:

https://github.com/simonbates/c2lc-coding-environment/commit/c0393d50c297746c8a870a67b45f82c4cdc4c259

I changed the logic to always scroll to index+1 on command override. When I was doing some manual testing, I found a case where the previous logic did not scroll when I think we should:

- load the coding environment (the following assumes that 5 steps are fully visible, and a 6th step is partially visible) 
- select the 'Forward' command
- set 5 'Forward's for steps 1-5

Expect:

- Scroll to see step 6

Actual:

- Does not scroll

Could you merge my change into your pull request?

simonbates:
Please update `ProgramBlockEditor.test.js` to use the new refactored structure.

PR #42, opened 2020-02-04
C2LC-158: Change layout of the ProgramBlockEditor Vertically

chosww:


chosww:
This PR has folded into C2LC-160

PR #43, opened 2020-02-04
C2LC-159: Draw lines in between steps

chosww:


chosww:
This PR has folded into C2LC-160

PR #48, opened 2020-02-18
C2LC-107: Disable the run button when program is empty

chosww:


simonbates:
This pull request does more than disable the run button when the program is empty. Please remove the additional functionality.

PR #49, opened 2020-02-28
C2LC-171: Implement a little node component between steps to insert a command between the steps

chosww:


src/AddNode.js
```diff
@@ -0,0 +1,90 @@
...
+import AriaDisablingButton from './AriaDisablingButton';
+import { ReactComponent as AddIcon } from './svg/Add.svg';
+import classNames from 'classnames';
+import './AddNode.scss';
+
+// TODO: I think if we don't need to programmatically focus AddNode instances,
+//       we can make AddNode a regular component rather than React.forwardRef
+// TODO: When we add a command to the end of a program using the keyboard,
+//       should focus remain on the final add node, or should it be on the
+//       newly inserted command?
```
simonbates, line 13: These TODO comments can be removed now.

src/AddNodeToggleSwitch.scss
```diff
@@ -0,0 +1,37 @@
+@import 'variables';
+
+.AddNodeToggleSwitch__switch {
```
simonbates, line 3: We can remove the "switch" Element part from the `AddNodeToggleSwitch__switch` class name as it refers to the whole component. `AddNodeToggleSwitch__switch-inner-circle` is good as it refers to a part of the component.

So:

- `AddNodeToggleSwitch__switch` -> `AddNodeToggleSwitch`
- `AddNodeToggleSwitch__switch--checked` -> `AddNodeToggleSwitch--checked`
- and so on

src/AddNodeToggleSwitch.js
```diff
@@ -0,0 +1,38 @@
...
+import React from 'react';
+import { injectIntl } from 'react-intl';
+import { ReactComponent as AddIcon } from './svg/Add.svg';
+import './AddNodeToggleSwitch.scss';
+
+type AddNodeToggleSwitchProps = {
+    intl: any,
+    isAddNodeExpandedMode: boolean,
+    onClick: () => void,
+    onKeyDown: (e: SyntheticKeyboardEvent<HTMLInputElement>) => void
```
simonbates, line 12: I suggest moving the click and keyboard handling from `ProgramBlockEditor` into `AddNodeToggleSwitch` and instead providing an `onChange` callback property. The user actions for toggling the control state are good candidates for being the responsibility of the AddNodeToggleSwitch control. This would enable us to move some code out of `ProgramBlockEditor` and put it closer to the UI for the control and in the future would make it easier to re-use the toggle control for other situations in the UI. Locating the user action logic in the component will ensure that all toggle switches operate the same and that we don't duplicate logic unnecessarily.

src/ProgramBlockEditor.js
```diff
@@ -174,22 +252,52 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
         return (
             <React.Fragment key={programStepNumber}>
                 <div className='ProgramBlockEditor__program-block-connector'/>
-                {this.makeProgramBlock(programStepNumber, command)}
+                {command !== 'addNode' ?
```
simonbates, line 255: Using a special command name here to select between (connector-AddNode-connector-tile) and (connector-AddNode) is unclear and fragile (for example, a user could create a custom tile called "addNode"). I would instead suggest making the difference explicit with separate functions:

- `makeProgramBlockSection` and
- `makeEndOfProgramAddNodeSection`

PR #57, opened 2020-05-21
C2LC-175: Update overall layout to not have page level scrollbar

chosww:


chosww:
We decided to have page level vertical scrollbar

PR #59, opened 2020-07-08
C2LC-192: Port co-design demo work

simonbates:
Add functionality developed to support C2LC co-design:

- Add audio playback  of sounds when buttons in the interface are clicked
- Run a program without a robot connected 
- Support drag and drop on the iPad using a polyfill from https://github.com/Bernardo-Castilho/dragdroptouch


src/CommandPaletteCommand.js
```diff
@@ -8,13 +8,15 @@ type CommandPaletteCommandProps = {
     commandName: string,
     intl: any,
     selectedCommandName: ?string,
+    audioManager: Object,
```
chosww, line 11: I remember I set audioManager's type to be Object, but According to flow documentation, "NOTE: For new code, prefer any or { [key: string]: any}. Object is an alias to any and will be deprecated and removed in a future version of Flow.". We should change its type to any or import AudioManger to CommandPaletteCommand then use it as type.

simonbates: Thanks Daniel! I've updated `CommandPaletteCommand` to use the `AudioManager` type.

PR #60, opened 2020-07-09
C2LC-193: Implement toggle switch to turn off/on audio feedback

chosww:


src/ProgramBlockEditor.js
```diff
@@ -362,9 +363,17 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
...
-                            isAddNodeExpandedMode={this.state.addNodeExpandedMode}
+                        <ToggleSwitch
+                            ariaLabel={this.props.intl.formatMessage({id:'ProgramBlockEditor.toggleAddNodeExpandMode'})}
+                            value={this.state.addNodeExpandedMode}
                             onChange={this.handleChangeAddNodeExpandedMode}
+                            contentsTrue={<AddIcon />}
+                            contentsFalse={<AddIcon />}
+                            className={
+                                this.state.addNodeExpandedMode ?
+                                    'ProgramBlockEditor__addNode ProgramBlockEditor__addNode--checked' :
```
simonbates, line 374: I think `ProgramBlockEditor__add-node-toggle-switch` would be better for the class name as this is for the switch and not for one of the add nodes themselves.

src/ToggleSwitch.js
```diff
@@ -0,0 +1,60 @@
...
+            this.props.className,
+            'ToggleSwitch'
+        );
+        return (
+            <div
+                className= {
+                    this.props.value ?
+                    `${classes} ToggleSwitch--checked` :
+                    classes
+                }
```
simonbates, line 44: We can include the `ToggleSwitch--checked` class name in the `classNames` call above:

    const classes = classNames(
        this.props.className,
        'ToggleSwitch',
        { 'ToggleSwitch--checked': this.props.value }
    );

And then in the JSX, `className` becomes just: `className={classes}`

PR #62, opened 2020-07-23
C2LC-210: Add condition check to handleDragStartCommand to enable drag…

chosww:


simonbates:
Thanks for the PR Daniel.

I would suggest tackling this is a slightly different manner. The `ProgramBlockEditor` has a property `editingDisabled` and I think we should base all of our checks regarding the editing of the program on this property. In this case: add a check to `ProgramBlockEditor` to ignore drop actions when `editingDisabled` is set to `true`. The `editingDisabled` property manages the editing enabled/disabled status of the editor and will be set in all cases that editing is disabled.

In your implementation for this PR, the check is done in `App` and only checks `state.interpreterIsRunning` -- this will mean that drag and drop will still work if editing is disabled for other reasons.

PR #63, opened 2020-07-27
C2LC-194: Add the Scene and the Robot

chosww:


src/Scene.js
```diff
@@ -90,37 +77,22 @@ export default class TurtleGraphics extends React.Component<{}, TurtleGraphicsSt
...
 
-    clear(): void {
-        this.setState({
-            path: []
-        });
-    }
-
     render() {
-        const turtleTransform = `translate(${this.state.location.x} ${this.state.location.y}) rotate(${this.state.directionDegrees} 0 0)`;
+        // Substract 90 degrees from the character bearing as the character image is drawn upright when it is facing East
```
simonbates, line 81: Typo: "Substract" should be "Subtract"

src/Scene.js
```diff
@@ -90,37 +77,22 @@ export default class TurtleGraphics extends React.Component<{}, TurtleGraphicsSt
...
-                                x2={pathSegment.x2}
-                                y2={pathSegment.y2}
-                                key={i} />
-                        })}
-                        <polygon
-                            className='TurtleGraphics__turtle'
-                            transform={turtleTransform}
-                            points='-6 4 6 4 0 -9'/>
+                        // Scene will have 9x5 squares with 42 unit
+                        viewBox='-189 -105 378 210'>
```
simonbates, line 94: As we discussed, let's use round numbers for the grid square size (100x100) and make the other dimensions (scene size, character size) relative to that.

src/Scene.scss
```diff
@@ -0,0 +1,14 @@
+@import 'variables';
+
+.Scene__drawing-area {
+    display: inline-block;
+    box-sizing: border-box;
+    width: 100%;
+    height: 55.5%;
```
simonbates, line 7: We can use `height: auto` here to maintain the aspect ratio of the SVG, without having to specify it explicitly in the CSS.

src/Scene.scss
```diff
@@ -0,0 +1,14 @@
...
+    border: 0.1rem solid #555;
+}
+
+.Scene__svg {
+    width: 100%;
+    height: 100%;
+}
```
simonbates, line 14: I think we don't need this `.Scene__svg` rule.

src/Scene.js
```diff
@@ -90,37 +77,22 @@ export default class TurtleGraphics extends React.Component<{}, TurtleGraphicsSt
...
+        // Substract 90 degrees from the character bearing as the character image is drawn upright when it is facing East
+        const robotCharacterTransform = `translate(${this.state.location.x} ${this.state.location.y}) rotate(${this.state.directionDegrees - 90} 0 0)`;
 
         return (
             <div>
                 <span
-                    className='TurtleGraphics__drawing-area'
+                    className='Scene__drawing-area'
                     role='img'
                     aria-label='Drawing area'>
```
simonbates, line 89: I know that we will be expanding on this in a later task, but for this PR, let's use "Scene" rather than "Drawing area". And let's move the text to our localized messages file.

src/Scene.js
```diff
@@ -90,37 +77,22 @@ export default class TurtleGraphics extends React.Component<{}, TurtleGraphicsSt
...
     render() {
-        const turtleTransform = `translate(${this.state.location.x} ${this.state.location.y}) rotate(${this.state.directionDegrees} 0 0)`;
+        // Substract 90 degrees from the character bearing as the character image is drawn upright when it is facing East
+        const robotCharacterTransform = `translate(${this.state.location.x} ${this.state.location.y}) rotate(${this.state.directionDegrees - 90} 0 0)`;
 
         return (
             <div>
                 <span
-                    className='TurtleGraphics__drawing-area'
+                    className='Scene__drawing-area'
```
simonbates, line 87: I think that this can just be "Scene", rather than "Scene__drawing-area"

src/_variables.scss
```diff
@@ -26,3 +26,9 @@ $c2lc-hover-indicator-width: 0.1rem;
 
 $c2lc-action-panel-width: 16rem;
 $c2lc-action-panel-height: 5rem;
+
+/* Scene */
+
+$c2lc-scene-tile-num-row: 5;
+$c2lc-scene-tile-num-column: 9;
+$c2lc-scene-tile-width: 100;
```
simonbates, line 34: As we are using "tile" for the program actions, I think using another term for these variable names would be clearer, such as "grid-cell".

src/App.scss
```diff
@@ -59,3 +59,7 @@ body {
 .App__command-palette-command {
     margin-top: 1rem;
 }
+
+:export {
+    characterMovementLength: $c2lc-scene-tile-width;
```
simonbates, line 64: This is a tricky one, but I think it's slightly preferable here to keep the existing name of `sceneGridCellWidth`, rather than introducing a new name.

simonbates:
The left over `home()` method on the Scene can be removed.

simonbates:
Please add tests for the Scene.

PR #67, opened 2020-07-30
C2LC-214: Fix NVDA focus issues

simonbates:
- Ensure that the tile element for which the ActionPanel is being
  opened does not move within the DOM by always including the
  'ProgramBlockEditor__action-panel-container-outer' element and adding
  the ActionPanel as a child of that

- Ensure that the final add node in the program editor is not
  re-rendered unnecessarily by giving it a unique stable React key
  ('endOfProgramAddNodeSection')

- Set focus when a program step is deleted (to either the step tile
  immediately following, if there is one, or the final add node)

This commit also does some refactoring and renaming to make the code
clearer.

### DIFF
--- a/src/ActionPanel.js
+++ b/src/ActionPanel.js
@@ -143,8 +143,6 @@ class ActionPanel extends React.Component<ActionPanelProps, {}> {
                 if(optionButtonRef) {
                     optionButtonRef.focus();
                 }
-            } else {
-                element.focus();
             }
         }
     }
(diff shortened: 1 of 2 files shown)


chosww:
Thanks Simon, I have tested the updated version on both Windows NVDA and Mac VoiceOver and I don't see the issue anymore. The changes to always have 'ProgramBlockEditor__action-panel-container-outer' div rendered in DOM and assigning unique key for the last AddNode make sense and they do minimize changes in DOM. Also, changing the focus logic when a step gets deleted works great and have a nice user experience.

PR #71, opened 2020-08-19
C2LC-195: Add a Grid to the Scene

chosww:


src/Scene.js
```diff
@@ -15,6 +15,61 @@ type SceneProps = {
 };
 
 class Scene extends React.Component<SceneProps, {}> {
+
+    drawGrid(numRow: number, numColumn: number, width: number, minX: number, minY: number) {
+        const grid = [];
+        let xOffset = minX;
```
simonbates, line 21: Within the block for the rows, `xOffset` never changes. The math would read more straightforwardly if we used `minX` directly, without the indirection to another variable.

And likewise for `yOffset` in the columns block.

src/Scene.js
```diff
@@ -15,6 +15,61 @@ type SceneProps = {
...
+        let yOffset = minY;
+        for (let i=1;i < numRow + 1;i++) {
+            yOffset = yOffset + width;
+            if (i < numRow) {
+            grid.push(<line
+                className='Scene__grid-line'
+                key={`grid-cell-row-${i}`}
+                x1={`${xOffset}`}
+                y1={`${yOffset}`}
+                x2={`${-xOffset}`}
```
simonbates, line 31: This calculation for the grid line endpoint assumes that the origin is in the middle. It is at the moment, but it will break if we change where the origin is (and because the calculation of the Scene bounds are in another function (`render`), we could break it without realizing there is a connection). The math could be generalized by passing in the Scene width and height (or calculating from the number of rows and columns and grid cell width), and using that.

Together with the suggestion above of using `minX` Instead of `xOffset`, we would use:

    x1={`${minX}`}
    y1={`${yOffset}`}
    x2={`${minX + sceneWidth}`}
    y2={`${yOffset}`}

src/Scene.js
```diff
@@ -15,6 +15,61 @@ type SceneProps = {
...
+                y2={`${yOffset}`} />);
+            }
+            grid.push(
+                <text
+                    className='Scene__grid-label'
+                    textAnchor='end'
+                    key={`grid-cell-label-${i}`}
+                    dominantBaseline='middle'
+                    x={`${xOffset * 1.05}`}
```
simonbates, line 40: This calculation also assumes that the origin is in the middle of the Scene. I would suggest instead calculating an offset relative to the width of the Scene (which is in effect what this calculation is doing, but do it more explicitly):

    minX - sceneWidth * 0.025

src/Scene.js
```diff
@@ -15,6 +15,61 @@ type SceneProps = {
...
+                    y={`${yOffset - width / 2}`}>
+                    {i}
+                </text>
+            )
+        }
+        xOffset = minX;
+        yOffset = minY;
+        for (let i=1;i < numColumn + 1;i++) {
```
simonbates, line 48: See above comments for the rows.

src/Scene.js
```diff
@@ -34,6 +89,12 @@ class Scene extends React.Component<SceneProps, {}> {
                     <svg
                         xmlns='http://www.w3.org/2000/svg'
                         viewBox={`${minX} ${minY} ${width} ${height}`}>
+                        <defs>
+                            <clipPath id='Scene'>
```
simonbates, line 93: Maybe `Scene-clippath` for the `id`.

simonbates:
Please add a test for the Scene. Test the rendering of the grid lines, labels, and `RobotCharacter` transform. No need to test with as many as 9x5 cells, we can test with a smaller number (1x1, 3x2) and also test some edge cases, such as 0 rows and/or columns.

simonbates:
The suggestions that I have made will decouple the grid drawing a little from the bounds calculation for the SVG (using `minX + sceneWidth`, rather than `-minX`). However, in the future we may want to decouple further to support panning of the Scene. We don't need to do this now, but it's worth keeping in mind as something that we might want to do.

simonbates:
As discussed on Skype, we can remove the wrapping `` `${}` `` from the SVG attribute expressions.

src/Scene.js
```diff
@@ -15,6 +15,64 @@ type SceneProps = {
...
+        for (let i=1;i < this.props.numRows + 1;i++) {
+            yOffset = yOffset + this.props.gridCellWidth;
+            if (i < this.props.numRows) {
+            grid.push(<line
+                className='Scene__grid-line'
+                key={`grid-cell-row-${i}`}
+                x1={minX}
+                y1={yOffset}
+                x2={minX + sceneWidth}
+                y2={yOffset} />);
```
simonbates, line 36: Indent one level as it's within an if statement

src/Scene.js
```diff
@@ -15,6 +15,64 @@ type SceneProps = {
...
+        for (let i=1;i < this.props.numColumns + 1;i++) {
+            xOffset = xOffset + this.props.gridCellWidth;
+            if (i < this.props.numColumns) {
+            grid.push(<line
+                className='Scene__grid-line'
+                key={`grid-cell-column-${i}`}
+                x1={xOffset}
+                y1={minY}
+                x2={xOffset}
+                y2={minY + sceneHeight} />);
```
simonbates, line 60: Indent one level as it's within an if statement

src/Scene.js
```diff
@@ -15,6 +15,64 @@ type SceneProps = {
...
 class Scene extends React.Component<SceneProps, {}> {
+
+    drawGrid(sceneWidth: number, sceneHeight: number, minX: number, minY: number) {
+        const grid = [];
+        const rowLabelOffset = sceneWidth * 0.025;
+        const columnLabelOffset = sceneHeight * 0.025;
+        let yOffset = minY;
+        if (this.props.numRows === 0 || this.props.numColumns === 0) {
+            return grid;
+        }
```
simonbates, line 26: We can move the `if (this.props.numRows === 0 || this.props.numColumns === 0)` test up earlier in the method, underneath

    const grid = [];

src/Scene.js
```diff
@@ -34,10 +92,18 @@ class Scene extends React.Component<SceneProps, {}> {
...
-                        <RobotCharacter
-                            robotCharacterTransform={robotCharacterTransform}
-                            width={this.props.gridCellWidth * 0.8}
-                        />
+                        <defs>
+                            <clipPath id='Scene-clippath'>
+                                <rect x={minX} y={minY} width={width} height={height} />
+                            </clipPath>
+                        </defs>
+                        {this.drawGrid(width, height, minX, minY)}
```
simonbates, line 100: Let's change the parameter order to match our use of `rect` above:

    {this.drawGrid(minX, minY, width, height)}

simonbates: And to match the order of the values in the `viewBox`

src/Scene.test.js
```diff
@@ -0,0 +1,92 @@
...
+
+function findGridLabels(sceneWrapper) {
+    return sceneWrapper.find('.Scene__grid-label');
+}
+
+function findCharacter(sceneWrapper) {
+    return sceneWrapper.find('.RobotCharacter');
+}
+
+describe('When the Scene renders', () => {
```
simonbates, line 62: As we discussed on Skype, please extend the tests to check the gird line and label coordinates and the position (transform) of the character.

PR #72, opened 2020-08-20
C2LC-197: Move the play controls to under the scene

chosww:


src/PlayButton.js
```diff
@@ -0,0 +1,36 @@
...
+    intl: any,
+    interpreterIsRunning: boolean,
+    runButtonDisabled: boolean,
+    onClickRunButton: () => void
+};
+
+class PlayButton extends React.Component<PlayButtonProps, {}> {
+    render() {
+        return (
+            <div className='PlayButton-container'>
```
simonbates, line 19: I think the `PlayButton-container` div and styling would be better located in the App because they are dependent on the location within a grid. Which is the responsibility of the App layout.

simonbates:
Could you try changing how we manage the border around the Scene and controls below it? Rather than make a 3/4 border on each separately, put the Scene and the controls in a container and put a border on that. I'll need to see what effect it has on the code, but I think it might be easier to manage.

simonbates:
A couple of notes on this PR:

- We are mixing the terms "play" and "run". For example, the new component is called "PlayButton", but it has a property "onClickRunButton". We should tighten up our use of language. (In the case of the properties on the PlayButton component, I think we can drop the qualification and just use "disabled" and "onClick" as what they refer to is clear from the context, but the mix of play/run is broader than this.)
- In the current implementation, the run button label includes the text of the program. For example: "Run current program forward right". This PR would remove the program from the label.

I've asked Lisa about these 2 topics in the Skype group chat and let's hold off on making changes until we have discussed with her.

src/PlayButton.js
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import AriaDisablingButton from './AriaDisablingButton';
+import { injectIntl } from 'react-intl';
+import { ReactComponent as PlayIcon } from './svg/Play.svg';
+import './PlayButton.scss';
+
+type PlayButtonProps = {
+    intl: any,
+    interpreterIsRunning: boolean,
+    runButtonDisabled: boolean,
```
simonbates, line 12: Rename `runButtonDisabled` to `disabled`

src/PlayButton.js
```diff
@@ -0,0 +1,34 @@
...
+    onClickRunButton: () => void
```
simonbates, line 13: Rename `onClickRunButton` to `onClick`

src/PlayButton.test.js
```diff
@@ -0,0 +1,89 @@
...
+                },
+                props
+            )
+        )
+    );
+
+    return wrapper;
+}
+
+function getRunButton(playButtonWrapper) {
```
simonbates, line 42: Rename to `getPlayButton()`

src/PlayButton.test.js
```diff
@@ -0,0 +1,89 @@
...
+    return playButtonWrapper.find(AriaDisablingButton)
+        .filter('.PlayButton');
+}
+
+describe('The Run button class is changed when the program is running', () => {
```
simonbates, line 47: There are many places in this file where the button is called the "run button". Replace with "play button".

simonbates:
I've just merged Tony's C2LC-221 branch in develop-0.4. Could you please merge develop-0.4 into this branch as it contains quite a few changes. Thanks.

PR #73, opened 2020-08-21
C2LC-199: Make the robot draw when moving on the scene

chosww:


src/CharacterState.js
```diff
@@ -6,38 +6,48 @@ export default class CharacterState {
     xPos: number; // Positive x is East
     yPos: number; // Positive y is South
     directionDegrees: number; // 0 is North, 90 is East
+    path: Array<any>;
```
simonbates, line 9: I think it would be worth adding a type for the path segments to `types.js`:

    export type PathSegment = {
        x1: number,
        y1: number,
        x2: number,
        y2: number
    };

And then defining `path` as `Array<PathSegment>`

src/CharacterState.test.js
```diff
@@ -7,18 +7,26 @@ import * as C2lcMath from './C2lcMath';
 //       with Flow than casting the expect() result to 'any'.
 
 expect.extend({
-    toHaveCharacterState(received, xPos, yPos, directionDegrees) {
+    toHaveCharacterState(received, xPos, yPos, directionDegrees, path) {
+        const isEqualPath = received.path[0] != null && path[0] != null ?
```
simonbates, line 11: Jest provides a utility function that does a deep equality check:

https://jestjs.io/docs/en/expect#thisequalsa-b

So, we can change the path test to:

    this.equals(received.path, path)

This will make the code simpler and also enable us to test paths with more than one segment.

src/CharacterState.test.js
```diff
@@ -39,25 +48,25 @@ expect.extend({
...
 
 test('Given the character is facing South, then Forward movement should move South', () => {
-    (expect(new CharacterState(0, 0, 180).forward(100)): any).toHaveCharacterState(0, 100, 180);
+    (expect(new CharacterState(0, 0, 180, []).forward(100)): any).toHaveCharacterState(0, 100, 180, [{x1: 0, y1: 0, x2: 0, y2: 100 }]);
 });
 
 test('Given the character is facing 30 degrees N of E, then Forward should move in that direction', () => {
-    (expect(new CharacterState(0, 0, 60).forward(100)): any).toHaveCharacterState(86.6025, -50, 60);
+    (expect(new CharacterState(0, 0, 60, []).forward(100)): any).toHaveCharacterState(86.6025, -50, 60, [{x1: 0, y1: 0, x2: 86.6025, y2: -50}]);
 });
```
simonbates, line 61: Add a test with 2 movements to verify we get a path with 2 segments.

src/CharacterState.js
simonbates: Actually, we probably don't need it to be in `types.js`. It's probably fine to make it local to the `CharacterState.js` file. Just add:

    type PathSegment = {
        x1: number,
        y1: number,
        x2: number,
        y2: number
    };

to `CharacterState.js`. The way we do for React component property types. I don't think we need the `PathSegment` type independent of `CharacterState`.

src/CharacterState.test.js
simonbates: Sorry. Using a general deep equality check isn't going to work of course, as we need to use approximate equality testing on the coordinates. I think we should write a function for comparing paths.

simonbates:
I've made a commit that adds the `PathSegment` type and a `pathEquals()` method for testing path equality:

https://github.com/simonbates/c2lc-coding-environment/commit/267e92b17bfb8f9fa466317bea9ba2a9345a47d4

simonbates:
This morning, I merged the grid into `develop-0.4` and I have merged `develop-0.4` into my copy of the `C2LC-199` branch. Additionally, I made a small change in `Scene.js`: to extract a `drawCharacterPath()` method to follow the pattern of `drawGrid()`. Could you merge my `C2LC-199` branch into this PR please?

Could you also add a test to `Scene.test.js` to check the drawing of the character path. We don't need to test the generation of the path as we are testing that in `CharacterState.test.js`. But rather, test that the path in the `CharacterState` is drawn. Test with an empty path, a path with 1 segment, and a path with 2 segments.

Thanks.

src/Scene.test.js
```diff
@@ -227,10 +231,55 @@ describe('When the robot character renders, transform should apply', () => {
...
-            characterState: new CharacterState(0, 90, 0)
+            characterState: new CharacterState(0, 90, 0, [])
         });
         const robotCharacter = findRobotCharacter(sceneWrapper);
         expect(robotCharacter.get(0).props.transform)
             .toBe('translate(0 90) rotate(-90 0 0)');
     });
 });
+
+describe('When the robot character moves, it will draw its moving path on the scene', () => {
```
simonbates, line 242: Please rename this test to "When the Character has a path, it is drawn on the Scene"

src/Scene.test.js
```diff
@@ -227,10 +231,55 @@ describe('When the robot character renders, transform should apply', () => {
...
+        });
+        const robotCharacterPath = findRobotCharacterPath(sceneWrapper);
+        expect(robotCharacterPath.length).toBe(1);
+        expect(robotCharacterPath.get(0).props.x1).toBe(100);
+        expect(robotCharacterPath.get(0).props.y1).toBe(200);
+        expect(robotCharacterPath.get(0).props.x2).toBe(300);
+        expect(robotCharacterPath.get(0).props.y2).toBe(400);
+    });
+
+    test('When there is two path segments', () => {
```
simonbates, line 265: "When there are two path segments"

PR #77, opened 2020-09-10
C2LC-205: Implement screen reader markup for the scene

chosww:
This work is still WIP


sbates-idrc:
Accomplished using an aria live region updated as the character is moved on the scene. Closing.

PR #78, opened 2020-09-22
C2LC-225: Update the Robot Icon

chosww:


src/RobotCharacter.js
```diff
@@ -11,13 +11,11 @@ type RobotCharacterProps = {
 
 export default class RobotCharacter extends React.Component<RobotCharacterProps, {}> {
     render() {
-        const characterRadius = this.props.width / 2;
         const characterWidth = this.props.width * 0.75;
         return (
             <g
```
simonbates, line 16: If we no longer have the circle, can we get rid of this grouping element?

src/RobotCharacter.js
```diff
@@ -11,13 +11,11 @@ type RobotCharacterProps = {
 
 export default class RobotCharacter extends React.Component<RobotCharacterProps, {}> {
     render() {
-        const characterRadius = this.props.width / 2;
         const characterWidth = this.props.width * 0.75;
```
simonbates, line 14: We can get rid of this calculation and use `this.props.width` directly for the size of the robot as we don't have the circle anymore.

src/RobotCharacter.js
chosww: Seems like applying transform directly to the icon is not working

simonbates: Ok, thanks for trying it.

src/RobotCharacter.js
chosww: Setting this.props.width as width and height gives this result
Screen Shot 2020-09-22 at 3 55 11 PM

simonbates: Did you also update the x and y?

chosww: Ah right, setting correct x and y set the robot character to the center; though, I remembered chat with Sepi, and she said "I think the smaller version looked better"

simonbates: In that case, if we want the robot smaller, let's control it with the 0.8 size factor in the Scene, rather than having 2 numbers that are multiplied.

simonbates: That is: change the 0.8 in Scene and remove this extra factor from RobotCharacter.

chosww: Sounds good :)

PR #80, opened 2020-09-23
C2LC-200: Add the pen down/pen up toggle switch for drawing

chosww:


simonbates:
Please add a test to `CharacterState.test.js` to verify that no path is added with `forward()` is called with `drawingEnabled` of `false`.

PR #82, opened 2020-09-24
C2LC-226: Added HTML report on flow type coverage.

the-t-in-rtf:
See [C2LC-226](https://issues.fluidproject.org/browse/C2LC-226) for details.

cc: @simonbates 

(I keep CCing you because I can't suggest a reviewer, I think I need to be added to a different group.)

### DIFF
--- a/.flowcoverage-all
+++ b/.flowcoverage-all
@@ -7,7 +7,7 @@
   "globIncludePatterns": [
     "src/**/*.js"
   ],
-  "outputDir": "coverage/flow",
+  "outputDir": "flow-coverage",
   "threshold": 1,
   "reportTypes": "html"
 }
--- a/.flowcoverage-all
+++ b/.flowcoverage-all
@@ -1,0 +1,14 @@
+{
+  "concurrentFiles": 1,
+  "globExcludePatterns": [
+    "node_modules/**",
+    "src/vendor/**"
+  ],
+  "globIncludePatterns": [
+    "src/**/*.js"
+  ],
+  "outputDir": "coverage/flow",
+  "threshold": 0,
+  "reportTypes": "html"
+}
+
--- a/.flowcoverage-all
+++ b/.flowcoverage-all
@@ -8,7 +8,7 @@
     "src/**/*.js"
   ],
   "outputDir": "coverage/flow",
-  "threshold": 0,
+  "threshold": 1,
   "reportTypes": "html"
 }
 
--- a/.flowcoverage-main-threshold
+++ b/.flowcoverage-main-threshold
@@ -8,7 +8,7 @@
   "globIncludePatterns": [
     "src/**/*.js"
   ],
-  "outputDir": "coverage/flow",
+  "outputDir": "flow-coverage-main",
   "threshold": 94,
   "reportTypes": "text"
 }
--- a/.flowcoverage-tests
+++ b/.flowcoverage-tests
@@ -3,7 +3,7 @@
   "globIncludePatterns": [
     "src/**/*.test.js"
   ],
-  "outputDir": "coverage/flow-tests",
+  "outputDir": "coverage/flow/tests",
   "threshold": 92,
   "reportTypes": "text"
 }
--- a/.flowcoverage-tests-threshold
+++ b/.flowcoverage-tests-threshold
@@ -3,7 +3,7 @@
   "globIncludePatterns": [
     "src/**/*.test.js"
   ],
-  "outputDir": "coverage/flow/tests",
+  "outputDir": "flow-coverage-tests",
   "threshold": 92,
   "reportTypes": "text"
 }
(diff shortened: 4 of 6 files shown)


the-t-in-rtf:
OK, the report works now, but I wanted to mention a few things for the review.

The report shows up in the existing `coverage` directory and is archived with the test coverage report.  `coverage/index.html` is the test coverage report, `coverage/flow/index.html` is the flow type coverage report.

With the test coverage, if the test coverage is too low, it blocks uploading the HTML report.  I can't really see a good way to unblock that.  However, as the flow type report and threshold checks are separate, we can ensure that if we have a flow type coverage failure, we can still look at the report.  You can see the ordering in the latest builds.

simonbates:
Thanks Tony.

I think the ordering of the steps in the CI build is good. In the future, if we find that we are getting build failures but really want to see the Flow coverage HTML report, we could move it earlier and archive it separately. But I think to start, it's nice to have a single coverage zip file.

simonbates:
I'd like to move the coverage output directories, so that output of separate reports is kept separate, so there's no risk of them interfering with one another.

We have `outputDir` set in `.flowcoverage-main` and `.flowcoverage-tests` that I think doesn't get used right now as we are only generating `text` reports from those files, which is going to `stdout`. But, I'd like to keep them separate in case that changes and we do add coverage output to those in the future.

Is it possible to change the output directory of the test coverage reporting? Or is that under control of `react-scripts`?

If it's possible to change the coverage directory for test reporting. Then, let's use something like:

- `coverage/tests`: Test coverage
- `coverage/flow`: Flow HTML reporting from `.flowcoverage-all`
- `coverage/flow-coverage-main`: `outputDir` for `.flowcoverage-main`
- `coverage/flow-coverage-tests`: `outputDir` for `.flowcoverage-tests`

If we can't change the test reporting output directory, then:

- `coverage`: Test coverage
- `flow-coverage`: Flow HTML reporting from `.flowcoverage-all`
- `flow-coverage-main`: `outputDir` for `.flowcoverage-main`
- `flow-coverage-tests`: `outputDir` for `.flowcoverage-tests`

simonbates:
Another thought to communicate the separate roles of the flow coverage config files: for the `.flowcoverage-main` and `flowcoverage-tests` files, we could put `threshold` or `summary` in their names to make their role clearer. Maybe try some options and see what you think works best.

simonbates:
Something like:

- `.flowcoverage-all-report`
- `.flowcoverage-main-threshold`
- `.flowcoverage-tests-threshold`

the-t-in-rtf:
There's an option, but we can't change it until we eject from CRA.  See output:

```
These options in your package.json Jest configuration are not currently supported by Create React App:

  • coverageDirectory

If you wish to override other Jest options, you need to eject from the default setup. You can do so by running npm run eject but remember that this is a one-way operation. You may also file an issue with Create React App to discuss supporting more options out of the box.
```

I will follow the other strategy, where `/coverage` is just for test coverage and we use another dir for flow type coverage.

the-t-in-rtf:
Forgot to remove the disallowed setting, fixed now.  This should be ready for further review.

simonbates:
Please add the `/flow-coverage` directory to `.gitignore`

simonbates:
Now that we are archiving the Flow coverage and test coverage reports separately, we could move the Flow coverage report generation and archiving up in the build order. I'm thinking to just after Flow check. Then we would get Flow coverage reporting even if there is a test failure.

```
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index 86ca0cb..82dfc19 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,6 +37,13 @@ jobs:
         run: npm run build
       - name: Flow
         run: npx flow check
+      - name: Flow Type Coverage (HTML report)
+        run: npx flow-coverage-report --config .flowcoverage-all
+      - name: Archive Flow Type Coverage Report
+        uses: actions/upload-artifact@v2
+        with:
+          name: Flow Type Coverage Report
+          path: flow-coverage/
       - name: Test
         run: npm test -- --verbose --coverage
       - name: Archive Test Code Coverage Report
@@ -44,13 +51,6 @@ jobs:
         with:
           name: Test Code Coverage Report
           path: coverage/
-      - name: Flow Type Coverage (HTML report)
-        run: npx flow-coverage-report --config .flowcoverage-all
-      - name: Archive Flow Type Coverage Report
-        uses: actions/upload-artifact@v2
-        with:
-          name: Flow Type Coverage Report
-          path: flow-coverage/
       - name: Flow Type Coverage (main coverage threshold)
         run: npx flow-coverage-report --config .flowcoverage-main-threshold
       - name: Flow Type Coverage (test coverage threshold)
```

the-t-in-rtf:
@simonbates, I moved the type checking report right below `npx flow check`.  I left the threshold checks after the tests, because I assumed that we'd like to see the tests results rather than fail for type coverage thresholds before the tests run.

PR #83, opened 2020-09-28
C2LC-229: Move addNodeExpandedMode to App and default to true

simonbates:
Move the addNodeExpandedMode state to the App, and set it to true by default.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -33,7 +33,8 @@ type AppContext = {
 };
 
 type AppSettings = {
-    language: string
+    language: string,
+    addNodeExpandedMode: boolean
 };
 
 type AppState = {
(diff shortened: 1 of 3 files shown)


simonbates:
@chosww and/or @the-t-in-rtf if you could take a look at this PR, that would be awesome.

At this morning's C2LC check-in, we decided to change the add node expanded state to true by default. And during the work for C2LC-224 (https://issues.fluidproject.org/browse/C2LC-224), we discussed moving the `addNodeExpandedMode` state up to the `App`. I went ahead and made both these changes.

Tony, this will conflict with your C2LC-224 changes. But it would simplify that PR by removing the need to click the toggle switch to change the add node expanded state, instead, we can specify the state as a prop. I am happy to update C2LC-224 if we merge this first.

chosww:
Thanks Simon, the changes to moving addNodeExpandedMode state up to the App looks great. I missed the discussion on making the decision that was made for C2LC-224, but I agree that we should manage the state in the App for the future work, where we are going to have a dropdown menu for the settings.

PR #86, opened 2020-10-08
C2LC-198: Add a Refresh Button

chosww:


src/App.js
```diff
@@ -354,11 +363,20 @@ export default class App extends React.Component<{}, AppState> {
...
+                                            this.state.interpreterIsRunning ||
+                                            programIsEmpty(this.state.program)}
                                         onClick={this.handleClickPlay}
                                     />
                                 </div>
+                                <div className='App__refreshButton-container'>
+                                    <RefreshButton
+                                        interpreterIsRunning={this.state.interpreterIsRunning}
+                                        disabled={
+                                            this.state.interpreterIsRunning
```
simonbates, line 375: Please place the disabled expression all on one line: `disabled={this.state.interpreterIsRunning}`

src/CharacterState.js
```diff
@@ -74,4 +74,13 @@ export default class CharacterState {
             this.path
         );
     }
+
+    refresh(): CharacterState {
```
simonbates, line 78: The refresh operation should reset all of the character state: place the character at the centre of the scene and reset the orientation.

The question then is: who should manage the reset state? We have 2 related parts of code:

1. The App initiates the CharacterState starting state: `new CharacterState(0, 0, 90, [])`
2. In this PR, you have added a `refresh()` operation to `CharacterState`

I'm thinking that the action of refreshing should be outside of the responsibility of the `CharacterState`: the `CharacterState` is the state of the character now. And that the initialization of the `CharacterState` and resetting should be the responsibility of `App`.

A suggestion for an alternative approach:

- Add a new instance member to `App`: `startingCharacterState = new CharacterState(0, 0, 90, [])`
- Initialize `characterState` to this value: `characterState: startingCharacterState`
- In `handleRefresh()`, reset the `characterState` to this initial value: `this.setState({characterState: startingCharacterState})`

simonbates:
Please update the layout so that the Play button is centred.

src/CharacterState.js
chosww: Thanks Simon for the review, and the suggestions you gave me. I think having refresh method within CharacterState and make the API available in App through the CharacterState instance have benefit when it comes to unit testing. With. the benefit, should we continue with the alternative approach?

chosww: I think including test cases to check if CharacterState has initial value after refresh is called is important, and making CharcterState component to handle refresh method allows us to test that behaviour. If we were to handle the behaviour in App, there's no easy way to test the handler, as the handler is passed down to RefreshButton component, and testing the behaviour would require integration testing of App and RefreshButton.

simonbates: But that is the case at the moment. We have no test for the `handleRefresh()` method of App:

    handleRefresh = () => {
        this.setState((state) => {
            return {
                characterState: state.characterState.refresh()
            };
        });
    }

We have a unit test for the `CharacterState.refresh()` method. Which we would no longer need in my proposed approach above.

simonbates: I'm not saying that we don't want to test. More that I don't think either approach has an advantage with regards to testing. Given our current testing infrastructure, we don't have testing for the handlers in App.

chosww: Yes, so I was suggesting let's keep the current implementation in order to test refresh method returns initial CharacterState. We don't currently have test cases for handleRefresh(), but what it does is setting the value that's returned from characterState.refresh(), and we are testing the behaviour of refresh method in the unit test of CharacterState. If you think testing refresh is unnecessary, your proposal is a better approach and I am happy to go with that.

simonbates: Where I am thinking we will go with this in the future is to have an object that collects/manages states for things in the world and their initial states to support resetting. We don't have this object yet, and in the short term App can provide this role. But I think such a world state manager object is the way to go rather than placing the responsibility in `CharacterState`. For the reason that the 'reset' state won't always be at the centre of the Scene. In the future, objects will be placed in the Scene and when we reset, we would want them to reset to that position, not to the centre. So, something has to be responsible for managing the 'reset' states. And I think that's better managed outside of `CharacterState` and `ChartacterState` works nicely as "the current state of the character" (possibly with some history for undoing). But that starting or reset state is outside of what it manages.

simonbates: The problem with doing the resetting in `CharacterState.refresh()` is that we now have 2 places where the state is managed:

1. In the App constructor, where we initialize the `CharacterState`
2. In `CharacterState.refresh()`

Those 2 pieces of code would need to be kept in sync without obvious relationship. We could remove this coupling by having the `CharacterState` carry around its reset state, but that feels like it's going to get messy in the future. I'd prefer to collect reset states in one place.

chosww: Ah, I understand your point now. I think when we have multiple characters on the scene, It would also be better to have a world state manager and having different characters initial positions stored in the manager to reset their positions. I will update this PR with the alternative approach.

src/RefreshButton.scss
```diff
@@ -0,0 +1,56 @@
...
+button.btn.RefreshButton.RefreshButton--pressed.RefreshButton--disabled:active:focus {
+    background-color: $c2lc-disabled-color;
+}
+
+button.btn.RefreshButton--pressed svg,
+button.btn.RefreshButton--pressed.RefreshButton--disabled svg {
+    stroke: #FFF;
+}
+
+button.btn.RefreshButton.RefreshButton__run-button--disabled:active:focus {
```
simonbates, line 53: I think this rule isn't needed. I couldn't see where we were using the `RefreshButton__run-button--disabled` class.

src/App.js
```diff
@@ -71,6 +73,9 @@ export default class App extends React.Component<{}, AppState> {
             bluetoothApiIsAvailable: FeatureDetection.bluetoothApiIsAvailable()
         };
 
+        // Begin facing East
+        this.startingCharacterState = new CharacterState(0, 0, 2, []);
+
         this.state = {
             program: [],
             characterState: new CharacterState(0, 0, 2, []), // Begin facing East
```
simonbates, line 81: `characterState: this.startingCharacterState`

src/RefreshButton.scss
```diff
@@ -0,0 +1,45 @@
...
+button.btn.RefreshButton--disabled {
+    background-color: #FFF;
+    cursor: default;
+}
+
+button.btn.RefreshButton svg {
+    stroke: #000;
+}
+
+button.btn.RefreshButton--disabled svg {
```
simonbates, line 17: This rule is duplicated below. Please organize the rules in groups. Such as all rules for disabled state together. It will make it easier to spot duplication.

PR #88, opened 2020-10-26
C2LC-240: Change movement logic to be constrained to the grid

simonbates:


chosww (approved):
Thanks Simon, the changes look great -- model of using numbers 0 to 7 to represent compass directions works nicely with the new commands for 0.5 release.

PR #89, opened 2020-10-26
C2LC-241: Add new movement and turning commands

chosww:


src/messages.json
```diff
@@ -14,20 +14,23 @@
...
-        "CommandInfo.right": "turn right",
-        "CommandInfo.previousStep.forward": "before step {previousStepNumber} forward",
-        "CommandInfo.previousStep.left": "before step {previousStepNumber} turn left",
-        "CommandInfo.previousStep.right": "before step {previousStepNumber} turn right",
-        "CommandInfo.nextStep.forward": "after Step {nextStepNumber} forward",
-        "CommandInfo.nextStep.left": "after step {nextStepNumber} turn left",
-        "CommandInfo.nextStep.right": "after step {nextStepNumber} turn right",
+        "Command.forward1": "forward 1 square",
+        "Command.forward2": "forward 2 squares",
+        "Command.forward3": "forward 3 square",
```
simonbates, line 19: "forward 3 squares"

src/messages.json
```diff
@@ -14,20 +14,23 @@
...
+        "Command.left180": "turn left 180 degrees",
+        "Command.right45": "turn right 45 degrees",
+        "Command.right90": "turn right 90 degrees",
+        "Command.right180": "turn right 180 degrees",
+        "CommandInfo.previousStep.forward": "before step {previousStepNumber} {command}",
+        "CommandInfo.previousStep.left": "before step {previousStepNumber} {command}",
+        "CommandInfo.previousStep.right": "before step {previousStepNumber} {command}",
+        "CommandInfo.nextStep.forward": "after Step {nextStepNumber} {command}",
+        "CommandInfo.nextStep.left": "after step {nextStepNumber} {command}",
+        "CommandInfo.nextStep.right": "after step {nextStepNumber} {command}",
```
simonbates, line 31: The "CommandInfo.previousStep.\*" and "CommandInfo.nextStep.\*" messages are all the same. We only need one message for each.

src/ActionPanel.js
```diff
@@ -46,25 +46,33 @@ class ActionPanel extends React.Component<ActionPanelProps, {}> {
...
                 );
         }
 
         if (this.props.pressedStepIndex > 0) {
             const prevStepName = this.props.program[this.props.pressedStepIndex - 1];
             ariaLabelObj['previousStepInfo'] =
                 this.props.intl.formatMessage(
-                    { id: `CommandInfo.previousStep.${prevStepName}`},
-                    { previousStepNumber: this.props.pressedStepIndex }
+                    { id: `CommandInfo.previousStep.${prevStepName.split(/([0-9]+)/)[0]}`},
```
simonbates, line 59: We don't need to do this `split()` as all the "CommandInfo.previousStep.\*" messages are the same. See the note under `messages.json`.

src/ActionPanel.js
```diff
@@ -46,25 +46,33 @@ class ActionPanel extends React.Component<ActionPanelProps, {}> {
...
                 );
         }
 
         if (this.props.pressedStepIndex < (this.props.program.length - 1)) {
             const nextStepName = this.props.program[this.props.pressedStepIndex + 1];
             ariaLabelObj['nextStepInfo'] =
                 this.props.intl.formatMessage(
-                    { id: `CommandInfo.nextStep.${nextStepName}`},
-                    { nextStepNumber: this.props.pressedStepIndex + 2}
+                    { id: `CommandInfo.nextStep.${nextStepName.split(/([0-9]+)/)[0]}`},
```
simonbates, line 71: We don't need to do this `split()` as all the "CommandInfo.nextStep.\*" messages are the same. See the note under `messages.json`.

src/App.js
```diff
@@ -92,71 +91,66 @@ export default class App extends React.Component<{}, AppState> {
...
-                    };
-                });
-                return new Promise((resolve, reject) => {
-                    setTimeout(() => {
-                        resolve();
-                    }, 1750);
-                });
-            }
-        );
+        for(let i=0; i<this.commandNames.length; i++) {
```
simonbates, line 102: I'm not keen on this code to register the command handlers. I would prefer to follow the straightforward no-logic code that we used before. It's more verbose, but less error-prone and brittle. It's likely that we will soon 1) change the format of the commands to use parameters and 2) add new commands. When we make either of those changes we will have to make sure they work with this logic, or remove this logic. This code is baking the wrong model into the codebase (that command names are parsed to interpret their meaning). And it will make the code brittle if is persists for long.

src/messages.json
```diff
@@ -48,10 +51,9 @@
         "ProgramBlockEditor.betweenBlocks": "Add selected action {command} between position {prevCommand} and position {postCommand}",
         "ProgramBlockEditor.lastBlock": "Add selected action {command} to the end of the program",
         "ProgramBlockEditor.blocks.noCommandSelected": "Make sure an action is selected",
-        "ProgramBlockEditor.command.forward": "Forward, position {index} of current program",
-        "ProgramBlockEditor.command.left": "Left, position {index} of current program",
-        "ProgramBlockEditor.command.none": "Empty block, position {index} of current program",
-        "ProgramBlockEditor.command.right": "Right, position {index} of current program",
+        "ProgramBlockEditor.command.forward": "{command}, position {index} of current program",
+        "ProgramBlockEditor.command.left": "{command}, position {index} of current program",
+        "ProgramBlockEditor.command.right": "{command}, position {index} of current program",
```
simonbates, line 56: The "ProgramBlockEditor.command.*" messages are all the same. We only need one.

src/ProgramBlockEditor.js
```diff
@@ -274,8 +274,11 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
             hasActionPanelControl && 'ProgramBlockEditor__program-block--pressed'
         );
         const ariaLabel = this.props.intl.formatMessage(
-            { id: `ProgramBlockEditor.command.${command}` },
-            { index: programStepNumber + 1 }
+            { id: `ProgramBlockEditor.command.${command.split(/([0-9]+)/)[0]}` },
```
simonbates, line 277: We don't need to do this `split()` as all the "ProgramBlockEditor.command.*" messages are the same. See the note under `messages.json`.

simonbates:
And now that we have the Refresh button merged, could you update this branch with the latest 0.5? Thanks.

src/App.js
chosww: Thank you Simon for reviewing this PR! I think that this logic helps us to reduce duplication of registering similar commands, and it's not too complex logic -- as you know, what it does is to loop through all the available command names and register handler for the commands. I know there's some logic involved in converting degree to the new direction value, and. splitting commands' name and value followed by, and I am thinking we could refactor them out to a separate file and write test cases for them to ensure logics are working as we expected. To your point 1, when we change the format of the commands to use parameters, we can directly use those parameters to the command registration handler, and that also helps us to simplify the logic by not have to use split values. For the point 2, when we add new commands, we can simply add new commands to the commandName property and update the switch statement with new commands, or if the new commands are existing, but different values, we don't have to change the switch statement at all.

simonbates: It does reduce duplication but it introduces a model for the meaning of the command names that is brittle: that we parse the command names to understand what they do. In the future, we could automate the registration of the command handlers, in the manner similar to what you do here, but with a richer, more flexible model of command properties. But I think that it is too early to make the right model as we know that our command structure will be changing. And I don't what to invest too much time getting the model right for something that will change.

simonbates: It also makes it hard to change the time delay for different commands, without adding more logic. As we discussed the possibility of doing with Tony earlier today. In effect what you have is a single command handler that parses its command name to see what it should do. We have a lot more flexibility if command handlers are actually different.

simonbates: The `switch` statement is executed for every command invocation.

chosww: Right, yeah it's too early for building a model that will be changed in the next release. Also, I agree with you that making individual command registration for each command make it easier for us in setting different delay values for different turning angles, I did not consider that into implementation of this model. I will go ahead and make individual handler registrations.

simonbates: Thanks Daniel. I know it's verbose but I think it's our best option in the short term.

chosww: No problem and thanks for your comments! 👍

simonbates: An idea to reduce some of our duplication: we could introduce a utility function for making delay Promises (to `Utils.js`).

So that we could replace the following lines:

    return new Promise((resolve, reject) => {
        setTimeout(() => {
            resolve();
        }, 1750);
    });

with a single line (and still keep full flexibility):

    return makeDelayPromise(1750);

PR #90, opened 2020-10-27
C2LC-234: Sonify the character position and actions.

the-t-in-rtf:
See [C2LC-234](https://issues.fluidproject.org/browse/C2LC-234) for details.

Please note, this is a draft because:

1. We need a working approach for issues with jsdom lacking the `AudioBuffer` class.
2. I need to add at least crude flow types for the Tone.js content.
3. Once point one is addressed, I need specific tests for the new functionality.


the-t-in-rtf:
The test coverage levels are a direct result of skipping `App.test.js`, if I checkout `develop-0.5` and just skip the rendering test, a lot of code is suddenly untested.  Here's a comparison, `develop-0.5` with all tests on the left, and with just `App.test.js` disabled on the right.

Screenshot 2020-10-29 at 10 53 19

So it looks like we'll continue to have build failures until we can get the App to render under JSDom again.

the-t-in-rtf:
I got the coverage back up by adding rendering tests for all the components that were only tested via `App.test.js`.  We should still figure that part out, but we had a mix of components with their own tests and components without, it seemed like a reasonable change.  I will work on cleaning up a bit, adding tests for the new sounds, and rerecording a slightly louder version of the bells.

the-t-in-rtf:
@simonbates, before I forget, in addition to the WAV files I've committed, I have the original untrimmed m4a recordings for all of these as well.  Should I commit those or just keep them somewhere handy?

simonbates:
> @simonbates, before I forget, in addition to the WAV files I've committed, I have the original untrimmed m4a recordings for all of these as well. Should I commit those or just keep them somewhere handy?

It would be nice to have everything together. How big are they?

simonbates:
I've expanded on the Flow types for Tone.js, adding in some of the class hierarchy https://github.com/simonbates/c2lc-coding-environment/commit/eeb5954e88ab90b14909866743fcb1bc5ff2dc95

the-t-in-rtf:
> It would be nice to have everything together. How big are they?

The WAV files and m4a files together are 15Mb.  I was wondering if there' s a convention we use for "originals", like an "assets" directory or something.  I could also put them in a shared folder on Google Drive.

the-t-in-rtf:
> I've expanded on the Flow types for Tone.js, adding in some of the class hierarchy [simonbates@eeb5954](https://github.com/simonbates/c2lc-coding-environment/commit/eeb5954e88ab90b14909866743fcb1bc5ff2dc95)

Thanks, I merged with your branch again.  I had to add a few more for classes I'm using in some new tests, I'll commit that shortly.

src/App.js
```diff
@@ -34,6 +34,10 @@ type AppContext = {
     bluetoothApiIsAvailable: boolean
 };
 
+type AppProps = {
+    movementDelay: number
```
simonbates, line 38: As we discussed briefly this morning during planning, I'd suggest moving the `movementDelay` variable to `AppSettings`, an instance of which is stored inside `AppState` under the `settings` key. This is where we are currently storing user configurable settings that we might want to persist in the future. Let's also add units to the variable name: `movementDelayMs`

src/App.js
```diff
@@ -107,134 +109,199 @@ export default class App extends React.Component<{}, AppState> {
             'forward1',
             'moveCharacter',
             () => {
-                this.audioManager.playSound('forward1');
+                // TODO: Enable announcements again.
+                // this.audioManager.playAnnouncement('forward1');
+
                 this.setState((state) => {
+                    const newCharacterState = state.characterState.forward(
```
simonbates, line 116: In the other command handlers below, you are placing the arguments to the new state call on one line, please update this call to be consistent.

src/App.js
```diff
@@ -107,134 +109,199 @@ export default class App extends React.Component<{}, AppState> {
...
         this.interpreter.addCommandHandler(
             'forward2',
             'moveCharacter',
             () => {
-                this.audioManager.playSound('forward2');
+                // TODO: Enable announcements again.
+                // this.audioManager.playAnnouncement('forward2');
+
                 this.setState((state) => {
+                    const finalNewCharacterState = state.characterState.forward(2, state.drawingEnabled);
```
simonbates, line 140: In most of the other command handlers, you are using `newCharacterState`. Please update the command handlers to use a consistent name.

src/App.js
```diff
@@ -107,134 +109,199 @@ export default class App extends React.Component<{}, AppState> {
...
-                        characterState: state.characterState.turnRight(1)
+                        characterState: newCharacterState
                     };
                 });
-                return Utils.makeDelayedPromise(1750);
+                return Utils.makeDelayedPromise(this.state.settings.movementDelayMs);
             }
         );
 
+        // TODO: Discuss changing this incrementally as I did with forward2 and forward3.
```
simonbates, line 264: Is this comment still relevant?

flow-typed/npm/tone_v14.x.x.js
```diff
@@ -0,0 +1,82 @@
...
+    //       https://tonejs.github.io/docs/14.7.58/type/Frequency
+    //       https://tonejs.github.io/docs/14.7.58/fn/Frequency
+    //       This "Frequency" is the type one, maybe call it FrequencyType?
+    declare type FrequencyType = string;
+
+    declare export class FrequencyClass {
+        toMidi: any
+    }
+
+    declare export function Frequency(frequency: number | string): FrequencyClass;
```
simonbates, line 12: Add comment: `// https://tonejs.github.io/docs/14.7.58/fn/Frequency`

flow-typed/npm/tone_v14.x.x.js
```diff
@@ -0,0 +1,82 @@
+declare module "tone" {
+    // TODO: There are two things called "Frequency" in Tone.js
+    //       https://tonejs.github.io/docs/14.7.58/type/Frequency
+    //       https://tonejs.github.io/docs/14.7.58/fn/Frequency
+    //       This "Frequency" is the type one, maybe call it FrequencyType?
+    declare type FrequencyType = string;
+
+    declare export class FrequencyClass {
```
simonbates, line 8: Add comment: `// https://tonejs.github.io/docs/14.7.58/FrequencyClass`

flow-typed/npm/tone_v14.x.x.js
```diff
@@ -0,0 +1,82 @@
+declare module "tone" {
+    // TODO: There are two things called "Frequency" in Tone.js
```
simonbates, line 2: Now that we have renamed the type below to "FrequencyType", we can remove the "TODO" from the comment and keep a note about using the name "FrequencyType" for the "Frequency" type and the name "Frequency" for the function.

flow-typed/npm/tone_v14.x.x.js
simonbates: I think it's worth keeping our Flow types as close to the TypeScript types as we can. We can change the `Frequency` function type to:

    declare export function Frequency(value: TimeValue | FrequencyType ): FrequencyClass;

We have already created the `TimeValue` and `FrequencyType` types and can use those directly to match the TypeScript definition.

src/AudioManager.test.js
```diff
@@ -0,0 +1,73 @@
...
+            const midiNote: number = Frequency(noteForState).toMidi();
+            maxPitch = Math.max(maxPitch, midiNote);
+            minPitch = Math.min(minPitch, midiNote);
+            expect(midiNote).toBeGreaterThanOrEqual(0);
+            expect(midiNote).toBeLessThanOrEqual(127);
+        }
+
+        // const pitchRange = maxPitch - minPitch;
+        // expect(pitchRange).toBeGreaterThanOrEqual(0);
+        // expect(pitchRange).toBeLessThanOrEqual(12);
```
simonbates, line 69: With these lines commented out, it looks to me like `maxPitch` and `minPitch` aren't being used.

src/App.js
the-t-in-rtf: Nope, I'll get rid of it.

src/AudioManager.test.js
the-t-in-rtf: Oops, I had to comment that out when I was fixing something.  I'll put it back.

flow-typed/npm/tone_v14.x.x.js
the-t-in-rtf: Done.

flow-typed/npm/tone_v14.x.x.js
the-t-in-rtf: Done.

flow-typed/npm/tone_v14.x.x.js
the-t-in-rtf: I made a go at rewording this.

the-t-in-rtf:
@simonbates I believe I've addressed your latest feedback.

src/App.js
the-t-in-rtf: Yes, that was left over from the previous approach to handling `forward` / `forward2` / `forward3`.  I standardised on `newCharacterState`, as there's no longer a need to distinguish the changes further.

src/App.js
the-t-in-rtf: Fixed.

PR #91, opened 2020-10-29
C2LC-235: Save and share work

chosww:


package.json
```diff
@@ -13,6 +13,7 @@
         "react-dom": "16.9.0",
         "react-intl": "3.3.0",
         "react-mic": "12.4.1",
+        "react-router-dom": "^5.2.0",
```
simonbates, line 16: I think we aren't using React Router any longer, is this right?

src/App.js
```diff
@@ -522,7 +526,31 @@ export default class App extends React.Component<{}, AppState> {
...
+                    console.log(`Error parsing program: ${programQuery}`);
+                    console.log(err.toString());
+                }
+            }
+        }
+    }
+
     componentDidUpdate(prevProps: {}, prevState: AppState) {
+        if (this.state.program !== prevState.program) {
+            const encodedURI = encodeURI(this.programSerializer.serialize(this.state.program));
```
simonbates, line 548: This is definitely the right general approach, but we should use `encodeURIComponent()` rather than `encodeURI()`. `encodeURIComponent()` encodes characters that have special meaning in URIs, such as "?", whereas `encodeURI()` does not. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent. I would also suggest renaming this variable from 'encodedURI' to 'encodedProgram'.

src/ProgramSerializer.js
```diff
@@ -0,0 +1,55 @@
...
+                    programURL += 'r45'
+                    break;
+                case ('right90') :
+                    programURL += 'r90'
+                    break;
+                case ('right180') :
+                    programURL += 'r180'
+                    break;
+                default:
+                    break;
```
simonbates, line 45: Throw an Error here with: "Unrecognized program command when serializing program: ${program[i]}"

src/App.js
```diff
@@ -522,7 +526,31 @@ export default class App extends React.Component<{}, AppState> {
         );
     }
 
+    componentDidMount() {
+        if (window.location.search != null) {
+            const query = new URLSearchParams(window.location.search);
+            const programQuery = query.get('program');
```
simonbates, line 532: Let's create a class to manage the URL params called `C2lcURLParams`:

- Basically a simple wrapper around `URLSearchParams`
- `constructor(query)`
    - Creates a `URLSearchParams` instance and stores in an instance var
    - Called like: `const params = new C2lcURLParams(window.location.search)`
- `getVersion()`
    - Returns `query.get('v')`
- `getProgram()`
    - Returns `query.get('p')`

src/App.js
```diff
@@ -522,7 +526,31 @@ export default class App extends React.Component<{}, AppState> {
...
+        }
+    }
+
     componentDidUpdate(prevProps: {}, prevState: AppState) {
+        if (this.state.program !== prevState.program) {
+            const encodedURI = encodeURI(this.programSerializer.serialize(this.state.program));
+            window.history.pushState(
+                {program: encodedURI},
+                '',
+                `?program=${encodedURI}`);
```
simonbates, line 552: Let's make a function for generating the URL that takes an unencoded version string and and unencoded program string and does the necessary encoding and build the URL.

simonbates:
Please revert your changes to `package.json` and `package-lock.json` now that you have removed the React Router dependency.

simonbates:
Please fix the warnings regarding unnecessary backslash escaping: https://github.com/codelearncreate/c2lc-coding-environment/pull/91/files#diff-6505611930dd506aa6b9341154ec200713a3afc95d2f540acb258ce4c1fffac0

simonbates:
I've expanded the tests a little for the ProgramSerializer:

https://github.com/simonbates/c2lc-coding-environment/commit/9fd301b95055bfd819781566eddf262afad15ffd

PR #92, opened 2020-10-30
C2LC-244: Provide aria label for the Scene component to describe position of the character

chosww:
I've implemented a logic to set class property for column lookup (convert character's position to corresponding column label); however, this logic assumes that the numColumns to be an odd number.


chosww:
Maybe we should just use similar way to get column position as what I did for the row position -- remove the column lookup property and just get a label with String.fromCharCode(yPos + Math.ceil(numColumns/2))

src/messages.json
```diff
@@ -55,7 +55,8 @@
         "ProgramTextEditor.programLabel": "Program:",
         "ProgramSpeedController.slider": "Program play speed",
         "AudioFeedbackToggleSwitch.audioFeedback": "Audio feedback",
-        "Scene": "Scene",
+        "Scene.inBound": "Scene, {numColumns} by {numRows} grid with a robot character at column {xPos}, row {yPos}",
```
simonbates, line 58: I think we will need to include the direction that the character is facing also.

Please rename this message to `Scene.inBounds`.

src/messages.json
```diff
@@ -55,7 +55,8 @@
...
+        "Scene.outBound": "Scene, {numColumns} by {numRows} grid with a robot character outside of the scene",
```
simonbates, line 59: I think we will need to include the direction that the character is facing and some indication of where they are with respect to the scene: which side(s) they are located at.

Please rename this message to `Scene.outOfBounds`.

simonbates:
I've left some comments regarding the details of the in bounds and out of bounds messages. I think we should start a conversation with Sepi about how to best describe direction and out of bounds location. Do we use degrees, or do we use compass directions (45 degrees vs NE)? Do we use compass directions or "left", "right", "above", "below" for the out of bounds location?

PR #93, opened 2020-11-02
C2LC-245: Fix replace command focus trap

chosww:


chosww:
I was thinking, should we include a test case to check if CommandPaletteCommand component has "command-palette-command" className to ensure that they are included in the focus trap?

src/ProgramBlockEditor.js
```diff
@@ -484,7 +484,7 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
             if (this.state.replaceIsActive) {
                 this.props.focusTrapManager.setFocusTrap(
                     this.handleCloseReplaceFocusTrap,
-                    ['.replace-action-button', '.App__command-palette-command button'],
+                    ['.replace-action-button', '.command-palette-command'],
```
simonbates, line 487: Thanks for catching this issue. When we added the new commands and changed the layout of the command palette, we removed the class that was being used to set the focus trap, is that right?

The problem was then that we thought that the class was used only for style, when in fact it was used for the focus trap also (or at least didn't consider that it might have a role beyond styling). It would be good to avoid hitting this again in the future. The first idea that comes to mind is to name the classes that we use explicitly with the prefix "focus-trap-" so it's clear what their purpose is. In this case: "focus-trap-replace-command". Each focus trap would get its own class and all elements that are part of that focus trap would be assigned that class.

That should help us in the future. If the focus trap related classes are identified explicitly as supporting the focus traps, it should be harder to accidentally delete them.

src/ProgramBlockEditor.js
```diff
@@ -484,7 +484,7 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
             if (this.state.replaceIsActive) {
                 this.props.focusTrapManager.setFocusTrap(
                     this.handleCloseReplaceFocusTrap,
-                    ['.replace-action-button', '.App__command-palette-command button'],
+                    ['.replace-action-button', '.focus-trap-replace-command'],
```
simonbates, line 487: Let's apply this naming consistently. Change ".replace-action-button" to ".focus-trap-replace-action-button". And also use this new naming convention on the other focus trap:

    ['.ProgramBlockEditor__program-block--pressed', '.ActionPanel__panel button'],

PR #95, opened 2020-11-06
C2LC-254: Avoid playing sounds before they are loaded.

the-t-in-rtf:
See [C2LC-254](https://issues.fluidproject.org/browse/C2LC-254) for details.

### DIFF
--- a/flow-typed/npm/tone_v14.x.x.js
+++ b/flow-typed/npm/tone_v14.x.x.js
@@ -53,7 +53,8 @@ declare module "tone" {
 
     // https://tonejs.github.io/docs/14.7.58/Player
     declare export class Player extends Source {
-        constructor(url: string): Player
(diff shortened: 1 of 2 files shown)


simonbates:
Thanks Tony. A couple of comments:

- I'm able to reproduce a similar error when triggering `playAnnouncement()`, so I think we need a check there also; it looks like Player also has a `loaded` property
    - To reproduce, click one of the commands in the command palette, such as a forward command
- You added the `loaded` check to `playSoundForCharacterState()`, which calls `playPitchedSample()`; I suggest either moving the check to `playPitchedSample()` or marking `playPitchedSample()` as internal so that we don't call it directly in the future and get the error again

the-t-in-rtf:
I moved the check for the sampler closer to the usage of `triggerAttackRelease`.  I also added a check for the announcements.

PR #97, opened 2020-11-06
C2LC-246: Add speed control to change the speed that the program runs

chosww:


src/Interpreter.js
```diff
@@ -17,6 +17,7 @@ export default class Interpreter {
     programCounter: number;
     memory: { [string]: any };
     isRunning: boolean;
+    timeDelayMs: number;
```
simonbates, line 20: I think `stepTimeMs` is maybe a better name now that this is managed inside the Interpreter.

src/Interpreter.js
```diff
@@ -17,6 +17,7 @@ export default class Interpreter {
...
     onRunningStateChange: (InterpreterRunningState) => void;
 
     constructor(onRunningStateChange: (InterpreterRunningState) => void) {
```
simonbates, line 23: Add `stepTimeMs` as a parameter to the constructor and use this to initialize `this.stepTimeMs`. So that the starting value can be passed from the outside (`App`).

src/Interpreter.js
```diff
@@ -26,6 +27,7 @@ export default class Interpreter {
         this.memory = {};
         this.isRunning = false;
         this.onRunningStateChange = onRunningStateChange;
+        this.timeDelayMs = 1000;
```
simonbates, line 30: Initialize from a constructor parameter.

src/Interpreter.js
```diff
@@ -112,7 +114,7 @@ export default class Interpreter {
     callCommandHandlers(handlers: Array<CommandHandler>): Promise<any> {
         const promises = [];
         for (const handler of handlers) {
```
simonbates, line 116: Grab the `stepTimeMs` before entering the handler loop:

    const stepTimeMs = this.stepTimeMs;

And use this value in the loop to ensure that every handler for this step is given the same value.

src/App.js
```diff
@@ -457,6 +446,12 @@ export default class App extends React.Component<{}, AppState> {
         });
     }
 
+    handleChangeProgramSpeed = (e: SyntheticInputEvent<HTMLInputElement>) => {
+        const speedLookUp = [2000, 1500, 1000, 500, 250];
+        const newMovementDelay = speedLookUp[parseInt(e.target.value) - 1];
+        this.interpreter.timeDelayMs = newMovementDelay;
```
simonbates, line 452: Add a method to Interpreter `setStepTime()`, rather than setting the property directly.

src/ProgramSpeedController.js
```diff
@@ -0,0 +1,36 @@
...
+import { Form } from 'react-bootstrap';
+import type { IntlShape } from 'react-intl';
+import { injectIntl } from 'react-intl';
+import { ReactComponent as SlowIcon } from './svg/Slow.svg';
+import { ReactComponent as FastIcon } from './svg/Fast.svg';
+import './ProgramSpeedController.scss';
+
+type ProgramSpeedControllerProps = {
+    intl: IntlShape,
+    onChange: (e: SyntheticInputEvent<HTMLInputElement>) => void
```
simonbates, line 13: I suggest moving the handling of the `HTMLInputElement` event to this component: give it a bit more responsibility.

- Add a property which is the list of values (in our case, App will provide the list `[2000, 1500, 1000, 500, 250]`)
- The `onChange` callback is called with one of the values provides from the list -- so it's called with 2000 for example, or 500, rather than the `HTMLInputElement`

So clients are isolated more from the implementation details of this component: they provide a list of values, and get notified when one of them has been selected.

src/ProgramSpeedController.js
```diff
@@ -0,0 +1,40 @@
...
+
+import React from 'react';
+import { Form } from 'react-bootstrap';
+import type { IntlShape } from 'react-intl';
+import { injectIntl } from 'react-intl';
+import { ReactComponent as SlowIcon } from './svg/Slow.svg';
+import { ReactComponent as FastIcon } from './svg/Fast.svg';
+import './ProgramSpeedController.scss';
+
+type ProgramSpeedControllerProps = {
```
simonbates, line 11: Please add a TODO comments:

    // TODO: The Form.Control in configured for exactly 5 steps but the values
    //       array param can be any length. Update the code to either constrain
    //       the values array to 5 elements, or to calculate the number of steps
    //       for the slider from the values array.

PR #98, opened 2020-11-10
C2LC-248: Initial Implementation of share button.

the-t-in-rtf:
See [C2LC-248](https://issues.fluidproject.org/browse/C2LC-248) for details.


the-t-in-rtf:
Quick notes on this:

First, I didn't even try to position the share button.  I'll merge with upstream once [https://issues.fluidproject.org/browse/C2LC-246](https://issues.fluidproject.org/browse/C2LC-246) is merged and put it next to the speed controls.

Second, I used a Modal for now, I wanted to discuss perhaps using a Toast instead.

Third, obviously I'll add tests, since we can read the clipboard using the same API and confirm that the value was set.

the-t-in-rtf:
@simonbates, as discussed I just committed my work in progress on new tests for the `ShareButton` component.  There are two failures I could use help with.

```
 FAIL  src/ShareButton.test.js
  ● displays modal on click

    TypeError: Cannot read property 'writeText' of undefined

      38 |         // Copy the URL to the clipboard, see:
      39 |         // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText
    > 40 |         navigator.clipboard.writeText(currentUrl).then(() => {

```

I made an attempt at mocking that part of the API, but somehow what I did isn't exposed to the component itself, i.e. my tests can access my fake `readText` method, but the component can't access the similar `writeText` method.

I also commented out the rendering tests, which fail because I don't seem to have wired up the i18n stuff correctly.  I'll commit that as well shortly, but here's the error:

```
    Invariant Violation: [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.

      37 | it('renders without crashing', () => {
      38 |     const div = document.createElement('div');
    > 39 |     ReactDOM.render(<ShareButton intl={intl}/>, div);
         |              ^
      40 |     ReactDOM.unmountComponentAtNode(div);
      41 | });
      42 | 
```

I can break these down and research them on my own, but I figured you might be able to at least point me in the right direction.

simonbates:
I've updated the tests for `ShareButton` in this commit: https://github.com/simonbates/c2lc-coding-environment/commit/4de7c929a862266736f3dc522bc1e1117736c509

Changes:

- Use the async form of Jest's `test()` for the tests which click on the Share button, to manage the use of Promises in accessing the clipboard
    - https://jestjs.io/docs/en/asynchronous
- Add a callback to `ShareButton` to receive notification when the modal is shown
    - This was necessary due to the use of Promises: we need to wait until the state has been changed before inspecting it
    - I don't love adding a property to `ShareButton` just to support testing, but it's the best I could think of without making larger changes to the structure of the code
- Update the test names to use Given, When, Then style
- I packaged your clipboard implementation up into a new class `FakeClipboard`, to make it easier to use in multiple tests; I also updated the names to match those in the Flow type definition for clipboard: https://github.com/facebook/flow/blob/f4a9d2634bbe7f4b417289d5bcfaaddc481d192d/lib/bom.js#L314-L318

the-t-in-rtf:
Thanks, @simonbates, for the fix and the links.  I copied the `it` style from the rendering tests I was reading through, the other syntax makes more sense to me.  Jest has a really wide range of strategies for handling async calls, that's good to know.

I still need to fix up the rendering test, I'll give that a go on my own a bit later today.

src/ShareButton.css
```diff
@@ -0,0 +1,5 @@
+.ShareButton {
+    margin-top: 1rem;
+    height: 2.25rem;
```
simonbates, line 3: I think we don't need the `margin-top` and `height` properties here.

src/ProgramBlockEditor.js
```diff
@@ -41,6 +41,7 @@ type ProgramBlockEditorProps = {
 
 type ProgramBlockEditorState = {
     showConfirmDeleteAll: boolean,
+    showShareComplete: boolean,
```
simonbates, line 44: It looks like the changes in `ProgramBlockEditor.js` are left over from previous work and can be reverted?

src/App.scss
```diff
@@ -88,15 +88,33 @@ body {
     grid-gap: 1rem;
 }
 
+.App__playAndShare-container {
+    height: 10rem;
+    margin-top: 1rem;
+    background-color: #F1F1F1;
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr;
+
```
simonbates, line 97: Extra blank line here

src/ShareButton.js
```diff
@@ -0,0 +1,79 @@
...
+import { injectIntl } from 'react-intl';
+import type { IntlShape } from 'react-intl';
+
+import ShareCompleteModal from './ShareCompleteModal';
+
+import './ShareButton.css';
+
+type ShareButtonProps = {
+    intl: IntlShape,
+    disabled: boolean,
```
simonbates, line 14: The `disabled` property is never used.

src/ShareButton.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+type ShareButtonState = {
+    showShareComplete: boolean
+}
+
+class ShareButton extends React.Component<ShareButtonProps, ShareButtonState> {
+    static defaultProps = {
+        disabled: false,
+        showShareComplete: false
```
simonbates, line 25: `showShareComplete` is a state value, rather than props.

src/ShareButton.test.js
```diff
@@ -0,0 +1,109 @@
...
+                },
+                props
+            )
+        )
+    );
+
+    return wrapper;
+}
+
+// TODO: Figure how to do this properly with the required intl infrastructure
```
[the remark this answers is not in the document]
the-t-in-rtf: Fixed.

src/ShareButton.css
the-t-in-rtf: Removed.

src/App.scss
the-t-in-rtf: Removed.

src/ShareButton.js
the-t-in-rtf: Removed.

src/ShareButton.js
the-t-in-rtf: Fixed.

src/ProgramBlockEditor.js
the-t-in-rtf: I reviewed the changes, what's left now seems appropriate (the new state variable).

PR #99, opened 2020-11-10
C2LC-247: Persist the contents of the scene in the URL

chosww:


src/CharacterStateSerializer.js
```diff
@@ -0,0 +1,151 @@
...
+        if (encodedPathSegments != null) {
+            for (let i=0; i<encodedPathSegments.length; i++) {
+                const x1 = this.characterToPositionLookUpTable[encodedPathSegments[i].charAt(0)];
+                const y1 = this.characterToPositionLookUpTable[encodedPathSegments[i].charAt(1)];
+                const x2 = this.characterToPositionLookUpTable[encodedPathSegments[i].charAt(2)];
+                const y2 = this.characterToPositionLookUpTable[encodedPathSegments[i].charAt(3)];
+                if (x1 != null && y1 != null && x2 != null && y2 != null) {
+                    path.push({ x1, y1, x2, y2 });
+                } else {
+                    throw Error(`Invalid path coordinates x1=${x1} y1=${y1} x2=${x2} y2=${y2}`);
```
simonbates, line 101: If we use the input values (`encodedPathSegments[i].charAt(0)` and such) in the error message, it will make the message more helpful.

src/CharacterStateSerializer.js
```diff
@@ -0,0 +1,151 @@
+// @flow
+
+import CharacterState from './CharacterState';
+
+export default class CharacterStateSerializer {
+    characterToPositionLookUpTable: {[string]: number};
+    constructor() {
+        this.characterToPositionLookUpTable = {
```
simonbates, line 8: I think it would be really good if we could implement this as a function, the inverse of `getAlphabetCoord()` that we test in a similar manner to the test for `getAlphabetCoord()` where we test at '0', 'a', 'A', 'z', and 'Z'.

src/CharacterStateSerializer.js
```diff
@@ -0,0 +1,151 @@
...
+        }
+        return pathParam;
+    }
+
+    deserialize(text: string): CharacterState {
+        const xPos = this.characterToPositionLookUpTable[text.charAt(0)];
+        const yPos = this.characterToPositionLookUpTable[text.charAt(1)];
+        const direction = this.getDirectionFromCoord(text.charAt(2));
+        if (xPos == null || yPos == null || direction == null ) {
+            throw Error(`Invalid character position xPos=${xPos}, yPos=${yPos}, direction=${direction}`);
```
simonbates, line 87: Similar to the note on the error for invalid path segment, if we use the input values in the message, it will make the message more helpful.

PR #100, opened 2020-11-12
C2LC-256: Refactor sounds to better distinguish speed and turn direction.

the-t-in-rtf:
See [C2LC-256](https://issues.fluidproject.org/browse/C2LC-256) for details.


the-t-in-rtf:
This pull has been updated with develop-0.7.  It had been a while, I forgot that this also uses different sounds to make the forward "speeds" distinguishable.  The current approach for "backward" sounds doesn't match, and only uses one sound for all three "speeds".

It won't take long to rectify this, but I wanted to let Sepideh hear the current state and discuss with the group.  We can:

1. Use a single sound for all forward commands.
2. Use a single sound for all backward commands.
3. Use a single sound for each turn, rather than three variations.
4. Create two more sounds for the backward commands (shouldn't take more than an hour).

the-t-in-rtf:
As discussed in our meeting yesterday, I updated all sounds.  I added sounds to distinguish backward1 from backward2 and backward3.  I also moved the existing 90 degree turn sounds to replace the previous 180 degree turn sounds, that were playing the pattern too fast to be pleasant.  I generated new 90 degree turn sounds between the tempo of the 45 and 180 degree turns.  I have tested each on its own and in comparison with its "partner" (i.e. forward1 vs. backward1, left90 vs. right90).  The match sounds better to me, so this is ready for @sbates-idrc to review further.

src/AudioManagerImpl.js
```diff
@@ -1,11 +1,159 @@
...
+            "C3": "C3.wav",
+            "C4": "C4.wav",
+            "C5": "C5.wav",
+            "C6": "C6.wav"
+        },
+        baseUrl: "/audio/backward/1/"
+    },
+    backward2: {
+        urls: {
+            "C1": "C1.wav",
```
sbates-idrc, line 98: `backward2` does not have a "C0" entry. Is this an issue?

the-t-in-rtf: I couldn't make the instrument I used play anything lower than C1.

the-t-in-rtf: I'll try to remember how I managed it for forwards, which uses the same patch.

the-t-in-rtf: OK, I had to remind myself of how I did this.  The arpeggiator built into the CS-80 (emulator) can easily create the pattern used for the forward sounds, you set the arpeggiator to "down", hold C1, and it alternates C1 and C0 (an octave lower).   Ideally you would set the arpeggiator to "up" and trigger C0, but since the CS-80 doesn't respond to anything lower than C1, I had to get creative.

I ended up using the pitch adjusting ribbon control on the CS-80, setting it to a range of 24 steps, which meant that it ranged from an octave lower (hard left) to an octave higher (hard right) than the played note.  Using that, I could hold the ribbon control hard left, hit C1 and it played C0 and then C1 as desired.

The patch gets a bit "crunchy" on the low end, but the sound is still audible, and we now have a complete set.

PR #103, opened 2020-11-16
C2LC-259: Fix layout issue in small device window size

chosww:


the-t-in-rtf:
See [C2LC-259](https://issues.fluidproject.org/browse/C2LC-259) for details.

the-t-in-rtf:
So, I tested this with a very narrow desktop browser window and my Android phone.  The layout is as promised and works well enough.

Not sure if we want to make a new ticket, but the "delete all" modal does not flow on a mobile device, the "delete" button is only partially visible (no text).

PR #105, opened 2020-11-23
C2LC-203: Add a Stop Button

chosww:


src/PlayButton.scss
```diff
@@ -2,6 +2,7 @@
 
 button.btn.PlayButton--play {
     margin-top: 0.5rem;
+    margin-right: 1.5rem;
```
simonbates, line 5: I'd like to move this style instead to `App.scss` by adding a class to `<PlayButton class="App__playButton"...` in App. The `margin-right` is a property of the desired positioning within the App, rather than an intrinsic part of the PlayButton. We haven't always been super careful about this distinction, but as we implement the new layout, I'd like to take care with which styles are fundamental to the component, and which are a function of the component's usage in a parent. Taking care with this will make it easier to make the layout dynamic and adapt to different device sizes.

PR #106, opened 2020-11-27
C2LC-262: Integrate space and forest themes

chosww:


src/App.js
```diff
@@ -632,20 +647,31 @@ export default class App extends React.Component<{}, AppState> {
...
             window.history.pushState(
                 {
                     p: serializedProgram,
-                    c: serializedCharacterState
+                    c: serializedCharacterState,
+                    t: this.state.settings.theme
                 },
                 '',
-                Utils.generateEncodedProgramURL('0.5', serializedProgram, serializedCharacterState));
+                Utils.generateEncodedProgramURL('0.5', this.state.settings.theme, serializedProgram, serializedCharacterState));
```
simonbates, line 661: We need to bump up the version number to 0.6.

src/C2lcURLParams.js
```diff
@@ -18,4 +20,12 @@ export default class C2lcURLParams {
     getCharacterState() {
         return this.urlSearchParams.get('c');
     }
+
+    getTheme(): ThemeNames {
```
simonbates, line 24: I think I'd prefer to keep the responsibility of C2lcURLParams class to extracting strings from the URL. If we parse the theme here, we have an asymmetry with regards to program and character state as we don't parse these here. Let's keep this class limited to returning strings. And do the parsing elsewhere (in App).

src/types.js
```diff
@@ -9,6 +9,8 @@ export type DeviceConnectionStatus = 'notConnected' | 'connecting' | 'connected'
 
 export type EditorMode = 'text' | 'block';
 
+export type ThemeNames = 'default' | 'forest' | 'space';
```
simonbates, line 12: I think just `ThemeName` in the singular here as it's the name of a type.

src/index.scss
```diff
@@ -31,6 +31,17 @@ body {
...
+    background-color: #49827A;
+    background-image: url('./svg/ForestBackground.svg');
+    background-size: 13%;
+}
+
+body.space-theme {
+    background-color: #171C1C;
+    background-image: url('./svg/SpaceBackground.svg');
+}
+
```
simonbates, line 44: Could these be moved to `themes.css`?

simonbates:
I've made some changes to C2LC-262 here: https://github.com/simonbates/c2lc-coding-environment/commit/c3d65078341d3b332cca69b788220e699c8228dd

Please merge these changes into this PR.

As we discussed, please also rename the "RobotCharacter" to "Character".

Thanks.

PR #107, opened 2020-11-27
C2LC-251: Persist the program between browser sessions

chosww:


chosww:
I included logic to load previous coding environment from local storage inside of a condition block window.location.search != null, because this way, we don't have to specify two separate cases for loading from the local storage; I think we should load from the local storage when loading the coding environment from URL fails, and that can happen when window.location.search returns null/undefined, or when programQuery or characterStateQuery returns null/undefined. However, I have checked that window.location.search always returns an empty value that passes the condition, even when we don't specify url queries (http://localhost:3000 still passes the condition window.location.search != null). So I am thinking we can include the local storage loading logic inside of the condition block, unless we don't want to load from local storage when programQuery or characterStateQuery returns null/undefiend.

PR #110, opened 2020-12-07
C2LC-243: Generate new announcement sounds.

the-t-in-rtf:
See [C2LC-243](https://issues.fluidproject.org/browse/C2LC-243) for details.  Also addresses [C2LC-219](https://issues.fluidproject.org/browse/C2LC-219).


the-t-in-rtf:
As we discussed in the chat, this pull is currently blocked until we can find a TTS voice with acceptable licensing terms (the built-in voices in OS X are only licensed for non-commercial use).

the-t-in-rtf:
As discussed in the ticket, I updated this pull to use the WebSpeech API and internationalised messages for announcements.

simonbates:
I tried this PR out this morning on the following:

- Linux Firefox
- iPad Safari
- Chromebook Chrome

I encountered some minor-ish issues:

- iPad Safari
  - When I triggered another announcement while one was active, it sometimes interrupted and sometimes waited for the existing one to complete
  - I think that there's more of a delay before the speech starts on the iPad than on the Chromebook or Firefox on Linux
- Chromebook Chrome
  - Sometimes the announcements would not play
    - I didn't figure out yet what circumstances cause this and I saw no errors in the Developer Tools Console
 
I will continue to test and see if I can figure out causes or reproduction steps.

the-t-in-rtf:
@simonbates, after a closer reading of the [WebSpeech API docs](https://wicg.github.io/speech-api/#tts-attributes), I suspect that I need to check both the `speaking` and `pending` attributes.  I will try that and commit if it doesn't introduce any issues.

the-t-in-rtf:
I tested the additional "pending" check with Chrome, Firefox, and Safari, it works as expected.  @simonbates, please give it a try on other platforms and let me know what you find.

the-t-in-rtf:
@simonbates, since this is all text now, I also wanted to discuss a small improvement here.  I was thinking of better distinguishing the three uses of each action name, i.e.:

1. When a command is added to the program.
2. When a command is selected in the control panel.
3. When a command is deselected in the control panel.

As it's much easier to add more text than to generate new sound files, I was thinking of breaking out something like "turn right 90 degrees" into the following for the above uses:

1. "added a 90 degree right turn"
2. "90 degree right turn selected"
3. "no command selected"

If you're OK with the general idea, I can start a chat in the general room and solicit input about the wording.

simonbates:
> If you're OK with the general idea, I can start a chat in the general room and solicit input about the wording.

This structure looks good to me. Thanks.

the-t-in-rtf:
@simonbates, @chosww and I were discussing moving the announcement that is played when a command is selected or deselected.  Currently it's handled by the individual command panel icon's click handler.  We were discussing moving this up to `App`, presumably [around here](https://github.com/codelearncreate/c2lc-coding-environment/blob/v0.3.0/src/App.js#L189).

The problem I have is that the audio manager's `playAnnouncement` function needs to receive the `intl` property created when we wrap things in `IntlProvider`.  I don't see a good way to make that available where we need it without rearranging things, perhaps to have an "outer" `App` class that does the wrapping in `IntlProvider` and an "inner" class (say, `InnerApp`) that is basically the same as our current `App`, but which has access to `this.props.intl` by virtual of the "outer" class having wrapped it in `IntlProvider`.

As both of you have much longer experience with `react-intl`, I am hoping that one of you can either suggest a better approach or confirm that a wrapper is appropriate.

the-t-in-rtf:
I split out the wordings as discussed in the group chat yesterday:

1. Selecting an action on the command panel announces "selected movement {command}".
2. Deselecting an action on the command panel announces "no movement selected".
3. Adding a command announces "added movement {movement}"
4. Deleting a command announces "deleted movement {movement}"
5. Replacing a command announces "replaced movement {oldMovement} with {newMovement}"

chosww:
Initially, using Intel wrapper to wrap App was an option we were thinking of, and I still think that's appropriate approach -- here's the JIRA issue about that: [C2LC-23](https://issues.fluidproject.org/browse/C2LC-23)

simonbates:
As @chosww mentioned above, we were considering this issue earlier in the development of the project as documented in [C2LC-23](https://issues.fluidproject.org/browse/C2LC-23). That issue outlines two approaches:

1. Have an outer wrapper that does the `IntlProvider`, as you suggest (in the JIRA, we used the names `IntlContainer` and `App`)
2. Organize our UI code such that we don't need an `intl` instance at the App level, and move localized UI to sub-components

At the time, we selected to go with option 2.

But, I do think in this case, it does make sense to put the option 1 structure in place so that App has access to an `intl` instance and can then pass that along to the `AudioManager`.

the-t-in-rtf:
@simonbates  and @chosww , I have committed my work in progress on moving the announcements to `App.js`.  There are three issues:

1. Although the content is eventually announced, the delay is way too long.
2. The initial state is (eventually) announced, which we probably don't want.
3. The command block tests need to be refactored and moved to `App.test.js`.

Unless we can address the speed issues, I think we may just want to live with having the command block handle the announcements.

simonbates:
@the-t-in-rtf I've made a change to the mechanism used to announce the App `selectedAction` state:

https://github.com/simonbates/c2lc-coding-environment/commit/060ccd1c75699a52bb7245b0e4bd9f1b6fad91d1

Locally (Firefox and Chrome on Linux) I wasn't seeing exactly the same behaviour that you describe (delayed announcements), but I was definitely seeing funky behaviour: the same announcement being repeated or announcements just not being made at all.

The code was hooking into `App.getSelectedCommandName()` to make the `selectedAction` announcement. This method is called many times during each component update (as it is called for each CommandPaletteCommand) and was resulting in lots of announcement requests being made to the AudioManager.

My commit moves the announcement of `selectedAction` to the `componentDidUpdate()` method. Please try this out and let me know if it addresses the issues that you were seeing.

the-t-in-rtf:
Thanks, @simonbates, that makes a lot of sense and resolves the issues I was seeing.  I will work to move the relevant command block tests to App.test.js, after that I think we're good for another round of review.

the-t-in-rtf:
I couldn't quite figure out how to trigger a state change from the outside and wait for the results.  I pushed my work in progress up to get input.

the-t-in-rtf:
I had to figure a way of "unwrapping" the i18n layer to get to the app, which I did.  The tests now work in the new location, @simonbates, this should be ready for review again.

the-t-in-rtf:
@simonbates, the merge conflict resolution was a bit bumpy, but I think this is in a good state to continue reviewing.

src/AudioManagerImpl.js
```diff
@@ -106,68 +67,55 @@ implements AudioManager {
...
-        (value, key) => {
-            const player = new Player(value);
-            player.toDestination();
-            this.announcementLookUpTable[key] = player;
-        },
-        );
-    }
-
-    playAnnouncement(soundName: AnnouncedSoundName) {
+    playAnnouncement(messageIdSuffix: string, intl: IntlShape, messagePayload: any) {
```
simonbates, line 110: Small mismatch with the type defined in `types.js`:

    playAnnouncement(messageIdSuffix: string, intl: IntlShape, messageOptions?: any)

`messagePayload` vs `messageOptions?`

src/AudioManagerImpl.js
```diff
@@ -76,27 +39,25 @@ export function getNoteForState (characterState: CharacterState) : string {
     return noteName;
 }
 
-export default class AudioManagerImpl
-implements AudioManager {
+export default class AudioManager {
```
simonbates, line 42: Why the change here to remove the `implements AudioManager`?

src/App.js
```diff
@@ -735,6 +740,20 @@ export default class App extends React.Component<{}, AppState> {
                 }
             }
         }
+        if (this.state.selectedAction !== prevState.selectedAction) {
+            const messagePayload = {};
+            const announcementKey = this.state.selectedAction !== null ?
+                "movementSelected" : "noMovementSelected";
+            if (this.state.selectedAction) {
```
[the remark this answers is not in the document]
the-t-in-rtf: I changed both to use the same truthy check.

src/AudioManagerImpl.js
the-t-in-rtf: Pretty sure I messed it up when removing the type we got rid of.  It's fixed now.

src/AudioManagerImpl.js
the-t-in-rtf: Fixed.

src/App.js
```diff
@@ -531,135 +540,131 @@ export default class App extends React.Component<{}, AppState> {
...
-                                            onChange={this.handleTogglePenDown}/>
-                                    </div>
-                                    <div className='App__refreshButton-container'>
-                                        <RefreshButton
-                                            disabled={this.state.runningState === 'running'}
-                                            onClick={this.handleRefresh}
-                                        />
-                                    </div>
+        return (<React.Fragment>
+            <div
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
the-t-in-rtf: I also merged with upstream and checked the "files changed" to make sure there were no other large indentation-related updates.

the-t-in-rtf:
@simonbates, I have addressed the feedback you left, this should be ready for further review.

PR #111, opened 2020-12-07
C2LC-266: Add common visual focus indication for ProgramSpeedController in different browsers

chosww:


the-t-in-rtf:
In Safari, the styles for the play button highlighting and speed controls are different:

Screenshot 2021-01-26 at 10 33 08
Screenshot 2021-01-26 at 10 33 22

Every other element that highlights with blue on focus has a dark background, and it does seem like neighbouring elements should match.

the-t-in-rtf:
In looking at #117, I noticed that:

> we use blue when:
>
> 1. The button itself is black on a light background.
> 2. The button is light on a black background.

Neither of those is quite true of the speed selector.  Whether or not you agree, it still seems like it should match the play button highlighting.

chosww:
@the-t-in-rtf yeah, I had chat with Sepideh before about this highlight colour and she seemed fine with it, but it is worth asking her about your concern. Though, the blue highlighting is different than the focus colour for the toggle switches, the speed controller uses primary colour.

sepidehshahi:
I agree with @the-t-in-rtf. if we can change the color of the speed controller to black to match the play button and other controls' focus state would be great.

PR #113, opened 2020-12-10
C2LC-273: Added announcement when program block is added using drag and drop.

the-t-in-rtf:
See [C2LC-273](https://issues.fluidproject.org/browse/C2LC-273) for details.


the-t-in-rtf:
As we discussed, this is a good one to look at after #114

PR #114, opened 2020-12-10
C2LC-274: Add visual indication of where to drag and drop commands.

the-t-in-rtf:
See [C2LC-274](https://issues.fluidproject.org/browse/C2LC-274) for details.


the-t-in-rtf:
The last failure was a coverage check.  Previously we didn't have enough lines of drag and drop related stuff to hit the threshold, but the new logic took it over the line.

the-t-in-rtf:
As we discussed, we need to move the dragover and drop handlers to the right container, currently you can drop on the delete all icon and other parts of the program block editor.  Only the highlighted square should respond to drag over and drop events.  This should be a simple matter of moving the existing handlers around.

the-t-in-rtf:
The droppable region now matches the highlighting.

the-t-in-rtf:
@simonbates, this now includes the colour suggested by Lisa and Sepi in their review of my work to date.

They had asked about making the dragged icon have a border, which I couldn't figure an easy way to do and filed as [C2LC-279](https://issues.fluidproject.org/browse/C2LC-279).  As far as I know, this should be ready for review.

the-t-in-rtf:
@simonbates, merged with upstream and fixed the type check error that broke the last build.  This should be ready for review.

the-t-in-rtf:
In testing in our last meeting, @simonbates pointed out that there is a lag in determining the closest add node.  We need to verify and consider ways to mitigate this.  We should also test on a range of devices.

Also, as I pointed in the meeting, I could still use help from @chosww to figure out how to get the dragged icon to be consistent, i.e. with a border and the same border in all browsers, and the same no mattter whether the command was already selected before you started the drag.

the-t-in-rtf:
Sepi mentioned also that it would be good to further distinguish which node has been added.  She suggested making the icon 110% of the normal size and making the size normal after a short period has passed.  We might also choose to have the dragged icon appear bigger, then shrink after it has been placed.  We should write that up as another issue.

the-t-in-rtf:
I filed [C2LC-283](https://issues.fluidproject.org/browse/C2LC-283) for the icon resizing Sepi suggested.

the-t-in-rtf:
I added logic to ensure that the "closest add node" calculation is only run every 100ms.  Previously I could see that it was triggered dozens of times even during a short drag.

simonbates:
I'm seeing some flickering of the add node highlight as I drag an action tile around the program area. I believe that this is being caused by `ProgramBlockEditor.handleDragLeaveOnProgramArea()` being called when drag enters the area of a child of the `ProgramBlockEditor__program-sequence-scroll-container`, such as the `ProgramBlockEditor__program-sequence` or an AddNode drop area. I've attached two screenshots showing the dragged tile sitting on the boundary of the above mentioned children. In the screenshots, the add node is not highlighted.

Changes to temporarily add the borders:

```
> git diff
diff --git a/src/AddNode.scss b/src/AddNode.scss
index db3394b..42ae3ff 100644
--- a/src/AddNode.scss
+++ b/src/AddNode.scss
@@ -66,6 +66,7 @@ button.btn.AddNode__expanded-button svg {
         + #{2 * $addNodeExpandedButtonBorderWidth}
         + #{2 * $addNodeExpandedButtonPadding});
     height: $addNodeDropAreaHeight;
+    border: 1px solid red;
 }
 
 .AddNode--is-dragging-command .AddNode__collapsed-drop-area {
diff --git a/src/ProgramBlockEditor.scss b/src/ProgramBlockEditor.scss
index 51684e3..3e811f4 100644
--- a/src/ProgramBlockEditor.scss
+++ b/src/ProgramBlockEditor.scss
@@ -144,6 +144,7 @@ button.btn.ProgramBlockEditor__program-deleteAll-button--pressed.ProgramBlockEdi
     align-items: center;
     padding-top: 2.5rem;
     padding-bottom: 2.5rem;
+    border: 1px solid red;
 }
 
 .ProgramBlockEditor__start-indicator {
```

C2LC-274 program-sequence

C2LC-274 AddNode

simonbates:
We'd want to only remove the highlight if really leaving the program area.

simonbates:
This PR leaves drag and drop handling still functional on the AddNode. I'm thinking that we should remove the drag and drop from that component, so that we have only one set of logic, in the ProgramBlockEditor, for determining the target AddNode and handling the drop.

simonbates:
I tested on an iPad. The new behaviour generally works well: the add nodes are highlighted and reaction time is good. However, I did find one issue (that I haven't been able to reproduce on my desktop browsers):

- Use "Delete All" to make an empty program
- Drag and drop an action: add node is highlighted as expected
- Tap on the program step that was added, and delete the step
- Drag an action to the program area

Expected:

- The add node to be highlighted

Actual:

- This time the add node is not highlighted

But highlighting works again once a step has been added.

We can investigate this further once the changes discussed above (handling of drag leave to children, and removal of drag and drop from AddNode) have been added.

the-t-in-rtf:
> * This time the add node is not highlighted

I can't reproduce this on Safari, Chrome, or Firefox, so I'll need help tracking this down.

the-t-in-rtf:
> I'm seeing some flickering of the add node highlight as I drag an action tile around the program area.

I see this and will work on some ideas.

the-t-in-rtf:
@simonbates, I have committed a fix for the "drag leave" flicker.

the-t-in-rtf:
As we discussed in our meeting, the only thing I'm aware of left here is the appearance of the icon when it's dragged, which is already a problem and not anything introduced by the work here.  I'll write that up as a separate issue.

the-t-in-rtf:
See [C2LC-284](https://issues.fluidproject.org/browse/C2LC-284) for the remaining icon work to be done.

simonbates:
With the current drag and drop implementation, it isn't possible to edit the program while the program is running, only when the program is stopped. This PR makes it possible to drop into the program while the program is running. Please make it consistent with the current drag and drop behaviour.

It's probably a good idea to merge `develop-0.6` into this PR also to verify the behaviour with pausing.

What we want:

- Program running
  - Add nodes expanded
    - Dragging onto a running program does not highlight add nodes as drop is not possible
  - Add nodes collapsed
    - Dragging onto a running program does not highlight or expand add nodes as drop is not possible
- Program is stopped or paused
  - Add nodes expanded
    - Dragging onto a running program highlights add nodes as drop is possible
  - Add nodes collapsed
    - Dragging onto a running program highlights and expands add nodes as drop is possible

src/AddNode.js
```diff
@@ -12,61 +12,36 @@ type AddNodeProps = {
...
         const addNodeClasses = classNames(
             props.isDraggingCommand && 'AddNode--is-dragging-command'
         );
 
-        if (props.expandedMode || (isDragOver && !props.disabled)) {
+        const isNearestDragNode = ( props.programStepNumber === props.closestAddNodeIndex);
+        if (props.expandedMode || isNearestDragNode || (isDragOver && !props.disabled)) {
             return (
                 <div className={addNodeClasses}>
                     <div className='AddNode__drop-area-container'>
```
simonbates, line 38: We should be able to go further with the simplification of the `AddNode` component: now that it is no longer responsible for drag and drop, we can remove the `AddNode__drop-area-container` divs (this one and the one below for the collapsed case). And we can remove the associated style rules from `AddNode.scss`.

src/AddNode.js
```diff
@@ -12,61 +12,36 @@ type AddNodeProps = {
     disabled: boolean,
     'aria-label': string,
     isDraggingCommand: boolean,
+    closestAddNodeIndex: number,
     onClick: (stepNumber: number) => void,
     onDrop: (stepNumber: number) => void
```
simonbates, line 17: This property is no longer needed.

src/ProgramBlockEditor.js
```diff
@@ -248,11 +278,65 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
         this.insertSelectedCommandIntoProgram(stepNumber);
     };
 
+    // TODO: Discuss removing this once we have a good way to test drag and drop.
     /* istanbul ignore next */
-    handleDropCommand = (stepNumber: number) => {
+    handleDropCommandOnAddNode = (stepNumber: number) => {
```
simonbates, line 283: This method is no longer needed.

src/ProgramBlockEditor.js
```diff
@@ -343,11 +427,12 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
                     expandedMode={this.props.addNodeExpandedMode}
                     isDraggingCommand={this.props.isDraggingCommand}
                     programStepNumber={programStepNumber}
+                    closestAddNodeIndex={this.state.closestAddNodeIndex}
                     disabled={
                         this.props.editingDisabled ||
                         (!this.commandIsSelected() && !this.props.isDraggingCommand)}
                     onClick={this.handleClickAddNode}
-                    onDrop={this.handleDropCommand}
+                    onDrop={this.handleDropCommandOnAddNode}
```
simonbates, line 435: The `onDrop` callback is no longer needed.

src/ProgramBlockEditor.js
```diff
@@ -382,11 +467,12 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
                     expandedMode={true}
                     isDraggingCommand={this.props.isDraggingCommand}
                     programStepNumber={programStepNumber}
+                    closestAddNodeIndex={this.state.closestAddNodeIndex}
                     disabled={
                         this.props.editingDisabled ||
                         (!this.commandIsSelected() && !this.props.isDraggingCommand)}
                     onClick={this.handleClickAddNode}
-                    onDrop={this.handleDropCommand}
+                    onDrop={this.handleDropCommandOnAddNode}
```
simonbates, line 475: The `onDrop` callback is no longer needed.

src/ProgramBlockEditor.js
```diff
@@ -420,7 +506,10 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
         contents.push(this.makeEndOfProgramAddNodeSection(this.props.program.length));
 
         return (
-            <div className='ProgramBlockEditor__container'>
+            <div
+                className='ProgramBlockEditor__container'
+                ref={this.programBlockEditorRef}
```
simonbates, line 511: This `ref` needs to be moved to the element on which the `onDragOver`, `onDragLeave`, and `onDrop` events are registered: `ProgramBlockEditor__program-sequence-scroll-container`. Otherwise the source of the `onDragLeave` events will not line up with the bounds calculation in `handleDragLeaveOnProgramArea()`. This results in the highlight remaining in place when it should not. For example, drag a block into the program area, and then leave with it still 'in hand' through the left hand side: the first add node will remain highlighted even though we have left the drop area.

simonbates:
On Firefox (Linux) and not on Chrome: if I drag into the program area and then leave quickly via the top or bottom of the program area, sometimes the add node highlight persists when it should have been removed. I've not been able to reproduce on Chrome. Maybe there's a circumstance under which the leave event isn't fired?

src/AddNode.scss
```diff
@@ -75,3 +75,7 @@ button.btn.AddNode__expanded-button svg {
     width: (2 * $addNodeDropMargin) + $widthOfCollapsedIcon;
     height: $addNodeDropAreaHeight;
 }
+
+.AddNode__expanded-button.isDragTarget {
```
simonbates, line 79: We're using BEM naming for our classes, please rename to `.AddNode__expanded-button--isDragTarget`

src/ProgramBlockEditor.scss
```diff
@@ -194,3 +194,7 @@ button.ProgramBlockEditor__program-block--active {
     left: #{-$c2lc-action-panel-width / 2};
     top: #{-$c2lc-action-panel-height * 1.3};
 }
+
+.ProgramBlockEditor__program-sequence-scroll-container.isDragging {
```
simonbates, line 198: We're using BEM naming for our classes, please rename to `.ProgramBlockEditor__program-sequence-scroll-container--isDragging`

src/ProgramBlockEditor.js
```diff
@@ -248,11 +278,65 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
...
+
+    // TODO: Discuss removing this once we have a good way to test drag and drop.
+    /* istanbul ignore next */
+    handleDragLeaveOnProgramArea = (event: DragEvent) => {
+        const myBounds = this.programBlockEditorRef.current.getBoundingClientRect();
+        // Dragging over child elements with implicit drag and drop results in flickering.  Check to confirm whether we've actually left the area.
+        if (event.clientX >= myBounds.left &&
+            event.clientX <= (myBounds.left + myBounds.width) &&
+            event.clientY >= myBounds.top &&
+            event.clientY <= (myBounds.top + myBounds.height)) {
```
simonbates, line 314: If we invert the logic, we won't need the empty `if` block.

src/ProgramBlockEditor.js
simonbates: Ah, looks like there's already a `ref` on `ProgramBlockEditor__program-sequence-scroll-container`:

    ref={this.programSequenceContainerRef}

I missed it due to the way the code is formatted.

We can use this `ref` in the leave event bounds calculation.

the-t-in-rtf:
> On Firefox (Linux) and not on Chrome: if I drag into the program area and then leave quickly via the top or bottom of the program area, sometimes the add node highlight persists when it should have been removed. I've not been able to reproduce on Chrome. Maybe there's a circumstance under which the leave event isn't fired?

I see this as well, I think it's related to the boundary checking.  I am reviewing your comment on that and trying some things.

src/ProgramBlockEditor.js
the-t-in-rtf: Whatever we use as the `ref`, IMO it should be on the next line to avoid further confusion.

src/ProgramBlockEditor.js
the-t-in-rtf: Done.

src/AddNode.js
the-t-in-rtf: Done.

src/ProgramBlockEditor.js
the-t-in-rtf: Done.

src/ProgramBlockEditor.scss
the-t-in-rtf: I fixed this one.

src/AddNode.scss
the-t-in-rtf: Still working on this one, I think I'm missing another location where the old `isDragging` is added/removed.

src/ProgramBlockEditor.js
the-t-in-rtf: Done.

src/AddNode.scss
the-t-in-rtf: i.e. if I change the class in the two places I've found, the styles stop working.

src/ProgramBlockEditor.js
simonbates: Yeah, agreed on the `ref` placement.

src/AddNode.js
the-t-in-rtf: Done.

src/ProgramBlockEditor.js
the-t-in-rtf: Done.

src/AddNode.scss
the-t-in-rtf: Maybe I can pair with @chosww briefly on this on Monday.

the-t-in-rtf:
I think the recent refactor broke the behaviour when add nodes are collapsed, i.e. I can only get the first and last nodes to expand/highlight/work.  This needs to be fixed.

the-t-in-rtf:
On the chats, Simon suggested looking at which element is triggering the leave event:

> Thanks for making the updates. With regards to the flicker vs highlight sticking on leave. I suspect that it might be possible that highlight sticking could still happen without the flicker fix (the boundary checking). If that isn't the case, that removing the boundary checking fixes the highlighting getting stuck, that suggests to me that we should be able to fix the highlight getting stuck by another form of checking, such as rather than testing mouse position, test which element the leave event was fired by.

the-t-in-rtf:
> I think my recent refactor broke the behaviour when add nodes are collapsed, i.e. I can only get the first and last nodes to expand/highlight/work. This needs to be fixed.

In the debugger I can clearly see that `getClientBoundingRects` returns a 0x0 rectangle positioned at x:0, y:0.  If no expanded nodes are visible, then they all are equally distant and the first wins.  If an expanded node such as the single add node at the end is visible, that wins because it is much closer than x:0 and y:0.  I will have to do some research.

the-t-in-rtf:
It turns out that contracted add nodes didn't actually set their `ref` attribute, which is why we ended up with detached nodes with empty bounding boxes.  If I add `ref` to the contracted branch in `AddNode.js`, the bounding boxes and detection work again.  With this fix I can no longer get the highlight to stick either, so we may have addressed both.

the-t-in-rtf:
@simonbates, this should be ready for review again.

src/AddNode.scss
the-t-in-rtf: This does still need addressing, AFAIK it's the only thing left.

the-t-in-rtf: Figured it out.  The updated BEM selector wasn't specific enough to trump the styles imposed on button.btn.X, I had to add `!important` to the rule.

simonbates:
I'm still seeing the highlight flicker on and off as I drag a block within the program area. For example, if I drag a block up and down through the items in the program area. I've attached animated gif below.

Peek 2021-02-01 15-29

the-t-in-rtf:
> I'm still seeing the highlight flicker on and off as I drag a block within the program area.

In an earlier change I disabled the boundary checks on the drag leave handler.  This fixed the stickiness but results in the flicker again.  I'll debug a bit and see if your suggestion about checking the event details helps.

the-t-in-rtf:
I read [this stackoverflow post](https://stackoverflow.com/questions/7110353/html5-dragleave-fired-when-hovering-a-child-element), which suggested using CSS to address the issue.  this works, but disables clicking on the add nodes.  I will adjust that now.

the-t-in-rtf:
I ended up using another approach from the same post, which really sounds like something @simonbates was trying to suggest early on, i.e. checking `event.relatedTarget` to see if it's one of the program block editor's children.  This fixes the "flicker".

The "sticky highlighting" is nearly there, it sticks if you exit the drag area using the top border.  I will refine further.

the-t-in-rtf:
It's less elegant now, I had to special-case the "upstairs" and "lefthand" neighbours, which are considered as being contained by the drop target.  It does work though, all four borders can now trigger the leave properly, and there's no more flicker.  I think this is ready for further review.

src/AddNode.scss
```diff
@@ -47,30 +47,6 @@ button.btn.AddNode__expanded-button svg {
...
-}
-
-.AddNode--is-dragging-command .AddNode__collapsed-drop-area {
-    top: (-$addNodeDropAreaHeight / 2) + ($heightOfCollapsedIcon / 2);
-    left: -$addNodeDropMargin;
-    position: absolute;
-    width: (2 * $addNodeDropMargin) + $widthOfCollapsedIcon;
-    height: $addNodeDropAreaHeight;
+.AddNode__expanded-button--isDragTarget {
+    box-shadow: 0 0 0 0.4rem #53A28C !important;
```
simonbates, line 51: Rather than using `!important`, please make the selector more specific:

    .AddNode__expanded-button.AddNode__expanded-button--isDragTarget {
        box-shadow: 0 0 0 0.4rem #53A28C;
    }

Using `!important` will make applying themes more difficult as we will have to deal with relative importance in addition to relative specificity of selectors.

src/ProgramBlockEditor.js
```diff
@@ -247,10 +277,59 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
...
+    // TODO: Discuss removing this once we have a good way to test drag and drop.
+    /* istanbul ignore next */
+    handleDragLeaveOnProgramArea = (event: DragEvent) => {
+        if (!this.props.editingDisabled) {
+            // Ignore anything that we "contain", except for our upstairs and lefthand neighbours.
+            if (!event.relatedTarget ||
+                // $FlowFixMe: Flow doesn't recognise the relatedTarget property.
+                event.relatedTarget.className === "ProgramBlockEditor__header" ||
+                event.relatedTarget.className === "ProgramBlockEditor__container" ||
+                !this.programBlockEditorRef.current.contains(event.relatedTarget)
```
simonbates, line 308: As I mentioned in a previous comment, you are testing here for containment inside `this.programBlockEditorRef`. This is not the element to which you have attached the drag event handlers. If you change to checking for containment inside `this.programSequenceContainerRef` instead, you can remove the special casing for `"ProgramBlockEditor__header"` and `"ProgramBlockEditor__container"`.

the-t-in-rtf: Sorry, I misread your last comment as explaining why I could continue doing something, I didn't catch the difference in the names.  It's fixed now.

src/AddNode.scss
the-t-in-rtf: Fixed.

simonbates:
I've pushed a couple of changes to my copy of the C2LC-274 branch:

- https://github.com/simonbates/c2lc-coding-environment/commit/9bc0caab3b84652599b6a1901bff07c290100a03
  - Fixes the iPad issue described in https://github.com/codelearncreate/c2lc-coding-environment/pull/114#issuecomment-762441881
  - This was caused by a CSS rule for focus taking precedence over the drag target rule
  - We didn't see this in other browsers because they set focus on drag start, whereas iPad Safari does not
- https://github.com/simonbates/c2lc-coding-environment/commit/e4e224d681a5a077c4c95288559bc8888dd16edb
  - Cleanup of a state value no longer used
 
I'm seeing one outstanding issue: the flickering is still present in Safari (on Mac and on iPad). The strategy of checking the containment of `event.relatedTarget` works on Chrome and Firefox, but does not work on Safari, where it is `null`. I found the following bug against WebKit: https://bugs.webkit.org/show_bug.cgi?id=66547

the-t-in-rtf:
@simonbates, I merged your fixes, thanks.  I also implemented your suggestion on the group chat (checking client bounds if `event.relatedTarget` isn't available).  I believe I've fixed the flicker on Safari.

PR #116, opened 2020-12-23
C2LC-201: Implement pause program step

chosww:


simonbates:
There's a case that isn't handled yet: deleting the remaining steps when the program is paused.

To reproduce:

- Make a program with 3 steps
- Run the program and pause when the 1st step is running
- The program will now be paused on the 2nd step
- Delete the 2nd step
- As expected the pause point will now move to the last step
- Now delete the last step
- Press the Play button

Expected:

- The program to run (deleting the last step should have stopped the running of the program, and pressing Play should have run the program from the start)

Actual:

- Nothing happens

In `App.componentDidUpdate()` , we have:

    if (this.state.programSequence !== prevState.programSequence) {
        if (this.state.programSequence.getProgramLength() === 0) {
            this.setState({ runningState: 'stopped' });
        }
    }

This handles the case where all steps have been deleted. We will also have to handle the case when the `programCounter` >= `programLength`. If I'm understanding the logic correctly, I think the fix is to change the above to:

    if (this.state.programSequence !== prevState.programSequence) {
        if (this.state.programSequence.getProgramLength() === 0) {
            // All steps have been deleted
            this.setState({ runningState: 'stopped' });
        } else if (this.state.programSequence.getProgramCounter()
                >= this.state.programSequence.getProgramLength()) {
            // All steps from the programCounter onward have been deleted
            this.setState({ runningState: 'stopped' });
        }
    }

chosww:
@simonbates thanks for catching that issue, and your comment totally makes sense. I have updated PR with your logic, and now there is no weird behaviour for the scenario you found.

src/ProgramBlockEditor.js
```diff
@@ -496,20 +498,20 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
             }
             this.focusAddNodeIndex = null;
         }
-        if (this.props.activeProgramStepNum != null) {
+        if (this.props.runningState === 'running') {
             // Take a snapshot of activeProgramStepNum that we know is non-null
             // (this is primarily to keep Flow happy as it doesn't know that
             // subsequent lines don't change it).
```
simonbates, line 504: This comment is no longer relevant and can be removed.

src/ProgramSequence.js
```diff
@@ -0,0 +1,103 @@
...
+        const program = this.program.slice();
+        program[index] = command;
+        return program;
+    };
+
+    delete(index: number): Program {
+        const program = this.program.slice();
+        program.splice(index, 1);
+        return program;
+    };
```
simonbates, line 102: Each one of these methods is called from exactly one other method:

- `swapPosition` from `swapStep`
- `insert` from `insertStep`
- `overwrite` from `overwrireStep`
- `delete` from `deleteStep`

I think we can just inline the contents of these methods into the ones that use them. I think having the extra layer of lower methods is adding unnecessary complexity.

These lower level methods are tested from `ProgramSequence.test.js` and we'll need to update the tests accordingly.

src/ProgramSequence.test.js
```diff
@@ -0,0 +1,128 @@
...
+    let programSequence = new ProgramSequence(['f1', 'f2', 'f3'], 1);
+    programSequence = programSequence.deleteStep(2);
+    expect(programSequence.getProgram()).toStrictEqual(['f1', 'f2']);
+    expect(programSequence.getProgramCounter()).toBe(1);
+    programSequence = programSequence.deleteStep(1);
+    expect(programSequence.getProgram()).toStrictEqual(['f1']);
+    expect(programSequence.getProgramCounter()).toBe(1);
+});
+
+test('insertStep should increment programCounter', () => {
```
simonbates, line 59: `insertStep()` should only increment the `programCounter` if the step is being inserted is at or before the `programCounter`.

src/ProgramSequence.js
```diff
@@ -0,0 +1,97 @@
...
+        const program = this.program.slice();
+        if (program[indexFrom] != null && program[indexTo] != null) {
+            const currentStep = program[indexFrom];
+            program[indexFrom] = program[indexTo];
+            program[indexTo] = currentStep;
+        }
+        return this.updateProgram(program);
+    }
+
+    swapPosition(indexFrom: number, indexTo: number): Program {
```
simonbates, line 86: Please also remove `swapPosition()` and update the associated tests.

src/ProgramSequence.test.js
```diff
@@ -0,0 +1,96 @@
+// @flow
+
+import ProgramSequence from './ProgramSequence';
+import type { Program } from './types';
+
```
simonbates, line 5: You have removed the previous tests for `overwrite()` but not replaced them with tests for `overwriteStep()`.

PR #117, opened 2021-01-11
C2LC-280: Update focus style of the add node toggle switch

chosww:


the-t-in-rtf:
As in #111, I was a little confused by the competing highlighting styles, but I guess we use blue when:

1. The button itself is black on a light background.
2. The button is light on a black background.

So in this particular case it makes sense to me that the add node expander and "clear all" buttons have different highlighting, and the proposed change works as expected.

PR #118, opened 2021-01-11
C2LC-282: Update button components to have consistent interaction styles

chosww:


the-t-in-rtf:
For the future, I'd appreciate it if you'd include a description, usually I just link to the associated issue.  It makes it easier to figure out the context of the changes, and then is there for future reference as well.

Something like this is fine:

See [C2LC-282](https://issues.fluidproject.org/browse/C2LC-282) for details.

the-t-in-rtf:
Sadly the ticket also lacks a description.  At a minimum, please summarise what you're changing and how I should be able to verify the change.

the-t-in-rtf:
A short discussion of the goal here would really help, but in the mean time I tried out the preview and the live site to try and understand.  My sense is:

1. Light coloured elements like the button to clear the scene or program were not previously consistent.
2. This pull makes them all use the style that previously the "delete all program steps" button used, i.e. green highlighting with a black border.

Is that the goal?

chosww:
@the-t-in-rtf you are right about that the issue should have some description. This work was to fix some button components' interaction styles to match others -- as you can see from the commit messages, PlayButton didn't have hover style, DeleteAll button didn't have disabled style, and RefreshButton didn't have hover style.

PR #119, opened 2021-01-13
C2LC-272: Make the highlighting of the running program active step more prominent

chosww:


the-t-in-rtf:
From [C2LC-272](https://issues.fluidproject.org/browse/C2LC-272):

> Sepideh suggested instead of using a stroke around the tiles, use a frame, that has a small gap between the active tile and the frame.
> 
> Suggested colour for the frame is either #53A28C or black.

the-t-in-rtf:
This works as described in the ticket in all browsers.  Have you run it by Sepi?  I kind of think the "gap" should be a little larger, but it would be good to let her see it and comment.

chosww:
Thanks @the-t-in-rtf, I have showed this to Sepideh before and she said it looks great.

simonbates:
@chosww Could you merge the latest develop-0.6 branch into this one please? I'd like to see how it interacts with the pausing indication. Thanks very much.

PR #122, opened 2021-02-04
C2LC-302: Create 'docs' directory.

the-t-in-rtf:
See [C2LC-302](https://issues.fluidproject.org/browse/C2LC-302) and [the associated google doc](https://docs.google.com/document/d/1YyeLVG6Sk0uguth7TplK4gsPUQWVJ_aFoYpXWQtLRnY/edit#) for details.


chosww:
Thanks @the-t-in-rtf, the changes look good, but CI failing from the Documentation lint you added. I was wondering if there's a configuration we can use to ignore checks in README.flow-typed.md, but I am unsure where the file is even located?

the-t-in-rtf:
@chosww, that's an old copy of the file that crept back in when I merged with upstream.  It was in the root of the repo,  but should have been replaced by docs/flow-typing.md.   I removed the outdated duplicate and the builds should pass now.

chosww:
This PR looks good for @sbates-idrc to review.

README.md
```diff
@@ -66,3 +66,4 @@ We would like to thank these amazing open source projects and libraries that we
 | [github.com/react-bootstrap/react-bootstrap](https://github.com/react-bootstrap/react-bootstrap) | MIT License |
 | [github.com/sass/node-sass](https://github.com/sass/node-sass) | MIT License |
 | [loading.io/](https://loading.io/) | [Loading.io Free License](https://loading.io/license/#free-license) |
+| [Tone.js][https://tonejs.github.io]| [MIT License](https://github.com/Tonejs/Tone.js/blob/dev/LICENSE.md)|
```
sbates-idrc, line 69: Please match the format of the lines for the other GitHub hosted projects:

```
| [github.com/Tonejs/Tone.js](https://github.com/Tonejs/Tone.js) | MIT License |
```

And move it up one line, between "github.com/sass/node-sass" and "loading.io".

docs/adr/README.md
```diff
@@ -0,0 +1,15 @@
...
+on how/whether to update our architecture. For more information on what an ADR is, check out:
+
+1. [Michael Nygard's blog post on the topic.](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions)
+2. [The ADR GitHub organisation](https://adr.github.io)
+
+We use [this template based on Nygard's blog post](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md)
+as the starting point for our ADRs.
+
+ADRs should be submitted as pull requests against this repository, which add a new ADR to this directory. ADRs are
+named based on the order in which they are approved. For example, the first ADR approved is named `ADR001.md`, the
```
sbates-idrc, line 13: I think maybe order that they are created might be easier to manage, rather than approved. So that we don't need to rename other ADRs when one is approved. On a more general note, do we still want to try using ADRs? I know it's something that we discussed, but haven't used yet.

docs/contributing.md
```diff
@@ -0,0 +1,262 @@
...
+            <ClickCounter timesClicked={this.state.secondComponentClicks} onClick={this.handleSecondComponentClick}>
+        </React.Fragment>);
+    }
+}
+
+ReactDOM.render(<TopDownManager />, document.getElementById('root'));
+```
+
+When a state variable is updated, react knows to re-render any components that use that value as a property. In this way
+we keep UI components simple and consolidate the logic regaarding variables. The above example also illustrates how we
```
sbates-idrc, line 145: Spelling: "regaarding"

docs/contributing.md
```diff
@@ -0,0 +1,262 @@
...
+
+```shell
+npx flow-coverage-report --config .flowcoverage-main-threshold
+```
+
+If you need to add types for libraries or other content, see [our Flow typing document](flow-typing.md)
+
+## Writing Tests
+
+Changes to this package are expected to be accompanied by tests, ideally that demonstrate that the change works as
```
sbates-idrc, line 183: I think "project" rather than "package".

docs/flow-typing.md
```diff
@@ -0,0 +1,19 @@
...
+npx flow-typed install enzyme@3
+npx flow-typed install jest@24
+npx flow-typed install react-bootstrap@0.32
+npx flow-typed install react-intl@2
+```
+
+Note that there have been local changes to the flow types for `react-intl` and `react-bootstrap`, those definitions
+cannot simply be regenerated or updated.
+
+We have manually created types for packages that do note have flow-typed definitions, like `soundex` and `Tone.js`.
```
sbates-idrc, line 18: We don't use `soundex` any longer.

the-t-in-rtf:
As we discussed in today's meeting, the changes for C2LC-303 (linting) brought in with this pull set the line length to 120, which wraps oddly on GitHub and other 80-char contexts.  The solutions are either not to limit the length, or to limit it to 80.  We discussed proceeding with 80 for now.

.markdownlint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "line_length": {
+        "line_length": 120,
```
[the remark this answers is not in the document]
the-t-in-rtf: We agreed to try 80 for now, I will make that change shortly and report back if there are issues.

docs/adr/README.md
the-t-in-rtf: Agreed on the naming, I'll change that.

I still think it's a good idea and think we should use an ADR  at least once to see how it fits.  I was thinking it might be something to try for the work I suggested on the chats, i.e. converting all actions to be a set of asynchronous events (move, play a sound, trigger an animation). I'd imagine we might end up wanting to move action definitions around, make classes for them, in short, change enough of the architecture that an ADR is warranted.

In retrospect I think the flow/typescript issue might have been a good fit, but I think we'll get a good sense of what is a good fit if we try it once.

PR #126, opened 2021-02-22
C2LC-290: Worlds menu.

the-t-in-rtf:
See [C2LC-290](https://issues.fluidproject.org/browse/C2LC-290) for details.


the-t-in-rtf:
This will not be ready to review until issues inherited from the work on C2LC-286 are resolved (the position of the character is not correct at the moment).

We also need to discuss where the menu should actually live.  I put it where it used to be until we can review and discuss.

chosww:
Hey Tony, could you please merge in the latest 286 into this work? Changes to move the heading to the top left corner have some updates to overall layout template, so it makes sense for me to give you feedback related to the layout work with the changes merged in; however, I will still leave some comments about the layout work you did to explain why overall layout in current build site is broken.

src/App.scss
```diff
@@ -73,84 +104,109 @@ body {
...
-    margin: 3rem auto 0 auto;
+    width: 6rem;
     display: grid;
-    grid-template-columns: repeat(3, 4rem);
+    grid-template-columns: repeat(2, 2.5rem);
     grid-gap: 1rem;
 }
 
-.App__playAndShare-container {
+.App__program-block-editor {
```
chosww, line 146: I see you have updated grid-column-end to include the width of the world selector component you have added, but we don't need to update grid-row-end for the program block editor, because there was no changes to the rows of overall layout. With this new grid-row-end value, ProgramBlockEditor and PlayAndShare container are fighting for the row 4 -- let's change value of 'grid-row-end' back to what it was before.

`
 grid-row-end: 4;
`

src/App.scss
```diff
@@ -73,84 +104,109 @@ body {
...
+    margin-right: 0.5rem;
+    grid-column-start: 3;
+    grid-column-end: 5;
+    grid-row-start: 3;
+    grid-row-end: 5;
+    height: auto;
+}
+
+.App__playAndShare-background {
```
chosww, line 155: Now that there's a new column, we need to update grid-column-end of this.

`grid-column-end: 6`

src/App.scss
```diff
@@ -73,84 +104,109 @@ body {
...
+    grid-column-end: 5;
+    grid-row-start: 4;
+    grid-row-end: 5;
     background-color: #F1F1F1;
-    height: 10rem;
-    margin-top: 1rem;
+    height: 7rem;
+}
+
+.App__playAndShare-container {
```
chosww, line 164: For the same reason as its background, this container's grid-column-end needs to be updated as well.

`grid-column-end: 5`

src/App.scss
```diff
@@ -73,84 +104,109 @@ body {
     margin-left: 50px;
 }
 
-.App__program-section {
-    margin-top: 1.5rem;
+.App__world-selector-container {
```
chosww, line 107: This may change after the latest 286 gets merged in, but world-selector container need same amount of margin-right as the container for ProgramBlockEditor. Let's add the margin.

`margin-right: 0.5rem`

src/WorldSelector.js
```diff
@@ -0,0 +1,85 @@
...
+    intl: IntlShape,
+    onSelect: (value: WorldName) => void
+};
+
+type WorldSelectorState = {
+    world: WorldName
+}
+
+class WorldSelector extends React.Component<WorldSelectorProps, WorldSelectorState> {
+    constructor (props) {
```
chosww, line 25: In this implementation, state for world is used to let App know that there was changes in world, so that world state in App gets updated. However, if we just update world state in App directly from onClick/keyDown events from WorldSelector, logics can be simplified and we don't need to store this state in two separate components. Also, world state is needed for other components as well, so keeping this state in one place, at the parent class (App) would be a nice refactoring.

chosww: I suggest removing WorldSelectorState type, constructor, as well as componentDidUpdate method and use WorldSelector as a UI component.

src/WorldSelector.js
```diff
@@ -0,0 +1,85 @@
...
+        return (
+            <div className='WorldSelector'>
+                <HeaderIcon
+                    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world'})}
+                    className="HeaderIcon"
+                />
+                <RobotIcon
+                    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world.default'})}
+                    className={"WorldIcon" + (this.state.world === 'default' ? " WorldIcon--selected" : "") }
+                    onClick={(e: Event) => { this.handleCharacterClick(e, 'default')}}
```
chosww, line 62: Using arrow function inside of render method is not recommended, because this will create a new function each time the component renders. We can simplify this by assigning value property to each Character buttons, and get this value from handleCharacterClick. 

`onClick={this.handleCharacterClick}`

src/WorldSelector.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    render() {
+        return (
+            <div className='WorldSelector'>
+                <HeaderIcon
+                    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world'})}
+                    className="HeaderIcon"
+                />
+                <RobotIcon
```
chosww, line 59: Let's wrap this icon with a button, and move all the properties in this SVG to it. Also, it would be nice to make this as a toggle button by using aria-pressed property, so users know if the character is selected or not. We also need value property, that's assigned to each character name, in this case, 'default'. Note that we need to use world property from App, with changes to removing state from this component. 

`<button className= {"WorldIcon" + (this.props.world === 'default' ? " WorldIcon--selected" : "") } aria-label={this.props.intl.formatMessage({id:'WorldSelector.world.default}} value='default' onClick={this.handleCharacterClick} onKeyDown={this.handleCharacterKeyDown}> 

`

src/WorldSelector.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    componentDidUpdate(prevProps: WorldSelectorProps, prevState: WorldSelectorState) {
+        if (this.state.world !== prevState.world) {
+            this.props.onSelect(this.state.world);
+        }
+    }
+
+    handleCharacterClick = (event: Event, worldName: WorldName) => {
```
chosww, line 38: With the suggestions I gave, the handler only needs event parameter, and let's get worldName from button's value property. Also, update directly to App's world state by calling onSelect property. 

`
 this.props.onSelect(event.currentTarget.value)
`

src/WorldSelector.js
```diff
@@ -0,0 +1,85 @@
...
+        event.preventDefault();
+        this.setState({ world: worldName });
+    }
+
+    handleCharacterKeyDown = (event: KeyboardEvent, worldName: WorldName) => {
+        const toggleKeys = [' ', 'Enter'];
+
+        if (toggleKeys.indexOf(event.key) !== -1) {
+            event.preventDefault();
+            this.setState({ world: worldName });
```
chosww, line 48: Same here, update world state directly with onSelect property. 

`
this.props.onSelect(event.currentTarget.value)
`

src/WorldSelector.js
```diff
@@ -0,0 +1,85 @@
...
+                <HeaderIcon
+                    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world'})}
+                    className="HeaderIcon"
+                />
+                <RobotIcon
+                    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world.default'})}
+                    className={"WorldIcon" + (this.state.world === 'default' ? " WorldIcon--selected" : "") }
+                    onClick={(e: Event) => { this.handleCharacterClick(e, 'default')}}
+                    onKeyDown={(e: KeyboardEvent) => { this.handleCharacterKeyDown(e, 'default')}}
+                    tabIndex={0}
```
chosww, line 64: Don't need tabIndex with button component, let's remove this.

src/WorldSelector.scss
```diff
@@ -0,0 +1,40 @@
...
+    border-radius: 5px;
+    margin-top: 0.8rem;
+    padding: 0.4rem;
+}
+
+.WorldSelector .WorldIcon--selected {
+    background: #88C3BC;
+}
+
+.WorldSelector button:focus {
```
chosww, line 34: Focus and hover styles are applied to button component, but these weren't applying because current WorldSelector is not a button.

src/App.js
```diff
@@ -578,103 +590,99 @@ export class App extends React.Component<AppProps, AppState> {
...
-                                        disabled={
-                                            this.state.runningState === 'stopped'
-                                            || this.state.runningState === 'stopRequested'}
-                                        onClick={this.handleClickStop}/>
                                 </div>
+                            </div>
+                        </div>
+                    </div>
+                    <div className="App__world-selector-container">
+                        <WorldSelector onSelect={this.handleChangeWorld} />
```
chosww, line 641: Let's pass world state to WorldSelector.

`
world={this.state.setting.world}
`

src/WorldSelector.js
the-t-in-rtf: Good idea, and done.

src/WorldSelector.js
the-t-in-rtf: Fixed.

src/WorldSelector.js
the-t-in-rtf: Makes sense, and done.

src/App.scss
the-t-in-rtf: Done.

src/App.scss
the-t-in-rtf: Fixed.

src/App.scss
the-t-in-rtf: Fixed.

src/App.scss
the-t-in-rtf: Right.  Fixed.

src/WorldSelector.js
the-t-in-rtf: Done.

src/WorldSelector.js
the-t-in-rtf: Done.

src/WorldSelector.js
the-t-in-rtf: Done.

src/WorldSelector.scss
the-t-in-rtf: I see them now.

src/App.js

the-t-in-rtf:
Thanks for the feedback, @chosww.  I updated with C2LC-286 and applied the changes you suggested.

I also noticed that the program block editor character was not using the world, and fixed that.

the-t-in-rtf:
This should be ready for further review.

src/WorldSelector.js
```diff
@@ -0,0 +1,81 @@
...
+                    className="HeaderIcon"
+                />
+                <button
+                    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world.default'})}
+                    aria-pressed={this.props.world === 'default'}
+                    className={"WorldIcon" + (this.props.world === 'default' ? " WorldIcon--selected" : "") }
+                    onClick={this.handleCharacterClick}
+                    onKeyDown={this.handleCharacterKeyDown}
+                    value="default"
+                >
```
chosww, line 53: As I talked about in Matrix chat, let's use AriaDisablingButton component for these buttons. We need to get condition to disable these buttons from App, let's name the prop disabled.

```javascript
<AriaDisablingButton
    aria-label={this.props.intl.formatMessage({id:'WorldSelector.world.default'})}
    aria-pressed={this.props.world === 'default'}
    className={"WorldIcon" + (this.props.world === 'default' ? " WorldIcon--selected" : "") }
    disabled={this.props.disabled}
    onClick={this.handleCharacterClick}
    onKeyDown={this.handleCharacterKeyDown}
    value="default"
>
```

src/App.scss
```diff
@@ -73,84 +89,135 @@ body {
...
+    grid-column-start: 4;
+    grid-column-end: 5;
+    grid-row-start: 1;
+    grid-row-end: 3;
+    margin-right: 0.5rem;
+    width: 6rem;
```
chosww, line 98: Let's apply the same amount of margin as the scene to the top. 

`
margin-top: 0.5rem;
`

chosww:
Thanks Tony for the updating the PR, changes you made looks good, just changing the buttons to AriaDisablingButtons would make this PR ready for Simon to review.

src/App.scss
the-t-in-rtf: Done.

src/WorldSelector.js
the-t-in-rtf: Sounds good, I converted all three and updated App.js to disable the world selector when the program is running.

the-t-in-rtf:
Thanks, @chosww.  I made the changes you suggested.

sbates-idrc:
@the-t-in-rtf please resolve conflicts with `develop-0.7` and then I will merge

PR #127, opened 2021-03-01
C2LC-289: Add new themes and theme picker.

the-t-in-rtf:
Add the new themes and theme picker.  See [C2LC-289](https://issues.fluidproject.org/browse/C2LC-289) for details.

Includes an initial effort at [C2LC-287](https://issues.fluidproject.org/browse/C2LC-287) as well.


src/ActionPanel.scss
```diff
@@ -1,57 +1,91 @@
 @import 'variables';
 
-// TODO: Rework the arrow implementation. For example, it is not quite
-//       centered with respect to the action panel.
-
-.ActionPanel__panel {
-    width: $c2lc-action-panel-width;
+.ActionPanel__background {
+    align-self: center;
```
chosww, line 4: align-self needs to be used in a flex or grid container; since this container is not one of them, this property has no effect. We can remove this line.

src/ActionPanel.scss
```diff
@@ -1,57 +1,91 @@
...
     height: $c2lc-action-panel-height;
```
chosww, line 5: This div has same height and width as the inner child, which makes setting height and width to this container to have no effect. Making this container to automatically calculate its height and width from inner contents is a good approach here. We can remove width and height.

src/ActionPanel.scss
```diff
@@ -1,57 +1,91 @@
...
-    content: '';
     position: absolute;
-    left: 45%;
-    top: 100%;
-    border-left: 0.7rem solid transparent;
-    border-top: 1rem solid #F1F1F1;
-    border-right: 0.7rem solid transparent;
+    top: -0.3rem;
+    width: $c2lc-action-panel-width;
     z-index: 1;
```
chosww, line 9: We have been avoiding using z-index unless it's absolutely necessary, because more z-index we have in our code, harder it will be to maintain. We don't need to use z-index here, if we remove z-index property from the inner container as well. That way, the rendering order will decide which items will appear on top of each other. Let's remove this z-index.

src/ActionPanel.scss
```diff
@@ -1,57 +1,91 @@
...
+    padding-left: 2rem;
+    padding-right: 2rem;
     position: absolute;
-    left: 43.8%;
-    top: 100%;
-    border-left: 0.9rem solid transparent;
-    border-top: 1.2rem solid #939595;
-    border-right: 0.9rem solid transparent;
+    width: $c2lc-action-panel-width;
+    z-index: 2;
```
chosww, line 30: Let's remove this z-index with the previous one.

src/ActionPanel.scss
```diff
@@ -1,57 +1,91 @@
...
 button.btn.ActionPanel__action-buttons {
-    background-color: #8F8F8F;
-    width: 3rem;
-    height: 3rem;
-    border-radius: 50%;
-    padding: 0.3rem;
+    background-color: #4C9990;
+    border: none;
+    border-radius: 50%;    
+    height: 2.1rem;
```
chosww, line 41: We have been ensuring that all interactable items are at least 45px by 45px to meet accessibility requirements, but I noticed that the new stop button is designed to be smaller than that as well. Could you please confirm with Sepideh about this? If she thinks we need to meet 45px requirement, height and width needs to be changed to 2.5rem.

src/App.scss
```diff
@@ -73,84 +90,140 @@ body {
...
+
+.App__commandpalette-heading-short {
+    display: none;
+}
+
+.App__command-palette-command-container {
+    width: 100%;
+    height: 100%;
+    padding: 1rem;
+    overflow-y: scroll;
```
chosww, line 121: I observed this weird behaviour for Light theme in chrome browser, where the movements panel's width changes from when light theme is initially selected and then hovering over to one of movements commands, or clicking to the theme selector or the audio toggle switch. It appears to be not calculating width of the movements panel with the scroll bar when light theme is initially selected, and then adjust width with the scroll bar after an interaction with the components in the panel. I tried to find ways to resolve this issue, but best one I found was to set webkit-scrollbar property on this container with value initial; however, this fix hides the scroll bar on light theme. 

```
.App__command-palette-command-container::-webkit-scrollbar {
    appearance: initial;
}
```

src/Themes.scss
```diff
@@ -0,0 +1,1392 @@
...
+    .App__header {
+        background-color: black;
+        color: white;
+    }
+
+    .ThemeSelector {
+        button.btn {
+            background-color: white;
+            color: black;
+
```
chosww, line 885: ThemeSelector for the contrast theme changes its size on hover, because border is applied on hover. We can prevent this by applying border to its base style with same background-color.

```
button.btn {
            background-color: white;
            color: black;
            border: 2px solid white;
```

src/App.js
```diff
@@ -748,15 +742,12 @@ export class App extends React.Component<AppProps, AppState> {
...
-        if (this.state.settings.theme !== prevState.settings.theme) {
-            if (document.body) {
-                if (this.state.settings.theme === 'default') {
-                    document.body.className = '';
-                } else {
-                    document.body.className = `${this.state.settings.theme}-theme`;
-                }
-            }
+
+        // As long as we have no underlying "default" theme styles, we need to always set this to ensure that the default theme is applied on startup.
```
chosww, line 746: Right place to set this up startup behaviours would be in ComponentDidMount.

chosww:
Thanks @the-t-in-rtf for the amazing work, this work definitely have a lot of changes. It seems like there's more issues to resolve based on Sepideh's feedback, but I gave an initial review on some css rules and bugs I found. Other than that, here are some style issues I noticed, that may be overlap with what Sepideh has already found:

- In the light and default themes, focus style for the refresh button is thiner than other controls
- In the light, dark, and grayscale themes, there's no focus style for the expanded add nodes
- In the light theme, there's no hover style for the theme selector
- In the light theme, paused step indication animation is not strong enough with forward commands
- In the dark theme, focus style for the refresh button is different than other controls
- In the grayscale theme, the refresh button disappears when playing a program
- In the grayscale theme, the refresh button has different style on hover
- In the default, light, dark themes, refresh button's background disappears because they are the same colour as the scene
- In the high contrast theme, there's no hover styles for the toggle switches
- In the high contrast theme, delete all icon disappears on hover 
- In the high contrast theme, there's no hover style for the stop button

chosww:
I just found one more issue, buttons in delete all modal have disappeared.

src/ActionPanel.scss
the-t-in-rtf: Done.

src/ActionPanel.scss
the-t-in-rtf: Done.

src/ActionPanel.scss
the-t-in-rtf: Good to know, and removed.

src/ActionPanel.scss
the-t-in-rtf: Removed.

src/Themes.scss
the-t-in-rtf: I followed a similar strategy and removed all the padding hints as well.  The size remains stable in all themes now.

src/App.js
the-t-in-rtf: This results in a short "blip" on refresh, where the default theme appears briefly and then is replaced.  I'll commit what I have so we can pair on it.

src/App.scss
the-t-in-rtf: We should discuss this in our next meeting, either we need a proper fix or we need to decided which side effect to live with.

the-t-in-rtf:
> Other than that, here are some style issues I noticed, that may be overlap with what Sepideh has already found:

I really appreciate the attention to detail, it's really helpful.  Since I did a lot based on Sepi's feedback on Friday, I went through the issues you raised point by point.  These appear to be fixed:

1. > In the light theme, there's no hover style for the theme selector
2. > In the grayscale theme, the refresh button disappears when playing a program
3. > In the grayscale theme, the refresh button has different style on hover
4. > In the default, light, dark themes, refresh button's background disappears because they are the same colour as the scene
5. > In the high contrast theme, delete all icon disappears on hover

I think this next one is fixed, as I did update the pause animation on Friday based on Sepi's feedback.  Please try it out and confirm if this addresses the concerns:

1. > In the light theme, paused step indication animation is not strong enough with forward commands

These I have confirmed as still needing to be fixed and will comment as I fix them:

1. > In the light and default themes, focus style for the refresh button is thinner than other controls
2. > In the light, dark, and grayscale themes, there's no focus style for the expanded add nodes
3. > In the dark theme, focus style for the refresh button is different than other controls
4. > In the high contrast theme, there's no hover styles for the toggle switches
5. > In the high contrast theme, there's no hover style for the stop button

the-t-in-rtf:
> In the light and default themes, focus style for the refresh button is thinner than other controls

I updated this to use the focus width variable instead of falling back to the react-bootstrap defaults.

the-t-in-rtf:
Screenshot 2021-03-22 at 11 34 59

I noticed that in the default and light themes, the focus outline appears behind the connector.  I don't have a good way to fix this now, adjusting `z-index` values didn't seem to help.  Any ideas, @chosww?

the-t-in-rtf:
> In the light, dark, and grayscale themes, there's no focus style for the expanded add nodes

I made a pass at this, but we should review with Sepi.

the-t-in-rtf:
> In the dark theme, focus style for the refresh button is different than other controls

Fixed.

the-t-in-rtf:
> In the high contrast theme, there's no hover style for the stop button

Fixed.

the-t-in-rtf:
> In the high contrast theme, there's no hover styles for the toggle switches.

I need to ask Sepi about this, as the hover style is only expressed in terms of the "on" state.

the-t-in-rtf:
> In the high contrast theme, there's no hover styles for the toggle switches

I made a pass at this, but need to review w/ Sepi.

the-t-in-rtf:
> I noticed that in the default and light themes, the focus outline appears behind the connector. I don't have a good way to fix this now, adjusting z-index values didn't seem to help

So, eventually I added `position: relative` to the two elements in question, and got this working.  After reviewing the inherited styles, I realised I could fix this by simply setting the add node to `z-index: auto` rather than letting it inherit `1`.

PR #128, opened 2021-03-04
C2LC-301: Add the Scene column and row headers

chosww:


chosww:
There's a known issue where there's no margin to the right of the scene area with its background, but that will be fixed when we merge in WorldSelector work.

the-t-in-rtf:
See [C2LC-301](https://issues.fluidproject.org/browse/C2LC-301) for context.

the-t-in-rtf:
So, I will comment in-context on the code changes, but I started by trying out the deployed preview.  Here are a few notes.

First, the pendown/up and scene reset controls are missing.  Is that coming later?

The scrolling when the character moves is smooth and consistent, I like that.  The scrolling with a mouse/keyboard is not as pretty, you can end up with a fraction of a row/column. displayed, which results in jagged little bits of the header outline.  I was wondering if it were possible to make scrolling only possible in even rows and columns.

The sounds need work as well, the position data sent to the AudioManager is used to pan the sound left or right depending on the position of the character.  Put on headphones and try creating a simple program to move forward (i.e. to the right) a few times.  The sounds should initially be panned hard left and gradually transition to the right as the character moves left.  Instead, the sound starts dead centre and pans to the left as the character moves right.

I suspect this may have been introduced when the starting position changed, but regardless it needs to be fixed.

the-t-in-rtf:
Looks like the sound issues were introduced a while ago, I filed that as [C2LC-306](https://issues.fluidproject.org/browse/C2LC-306).

the-t-in-rtf:
I'd like to look through the code again once C2LC-286 is merged, but nothing stuck out to me on first review.

the-t-in-rtf:
> Looks like the sound issues were introduced a while ago, I filed that as [C2LC-306](https://issues.fluidproject.org/browse/C2LC-306).

Sigh, new headphones with no clear labels, the sounds are fine. 🤦

the-t-in-rtf:
> First, the pendown/up and scene reset controls are missing. Is that coming later?

We discussed in chat and this is now fixed.

We will review the scrolling and discuss options in our next meeting with the team.

PR #130, opened 2021-03-15
C2LC-252: Specify the home position for the character

chosww:


the-t-in-rtf:
See [C2LC-252](https://issues.fluidproject.org/browse/C2LC-252) for details.

the-t-in-rtf:
As usual I started by trying out the preview site.  A few observations:

First, if you move the character to a corner, point the character diagonally at the corner, and move forward, a very short path is drawn, basically a dot.  I'm not sure this is desirable.  We should discuss it with Sepi and others.

The arrows and turns work well, and the edge behaviour works well.  The only concern I have with the controls to manually input the value of the row or column.  A change in value doesn't take effect until you leave the input.  I would expect a carriage return to immediately attempt to update the value, as you would with a search form or other text input.

This lack of support for carriage returns makes it more confusing when an invalid value is entered.  I remember the discussions about changing the value back to the last valid value, but taken together, it's possible to enter an invalid row or column and not realise that it's invalid until you tab away.

I think adding an update on carriage return would be enough, as then the valid would "snap" back to the last value as a kind of immediate feedback.

src/Character.js
```diff
@@ -47,10 +55,25 @@ export default class Character extends React.Component<CharacterProps, {}> {
...
                 className='Character'
                 transform={this.props.transform}>
                 {this.getThemedCharacter()}
             </g>
         );
     }
+
+    componentDidMount() {
+        if (this.characterRef.current.scrollIntoView) {
+            this.characterRef.current.scrollIntoView({ behavior: 'auto', block: 'nearest', inline: 'nearest' });
```
the-t-in-rtf, line 68: If we never use different options, we should find a way to reuse them cleanly, so that they can't get out of sync if we need to change anything.  Defining a constant or creating a small helper function would be fine.

src/Character.test.js
```diff
@@ -49,3 +49,30 @@ describe('Right character should render based on theme props', () => {
...
+    const mockScrollIntoView = jest.fn();
+
+    window.HTMLElement.prototype.scrollIntoView = mockScrollIntoView;
+
+    const wrapper = createMountCharacter();
+
+    // When the component mounts, call scrollIntoView
+    expect(mockScrollIntoView.mock.calls.length).toBe(1);
+    expect(mockScrollIntoView.mock.calls[0][0]).toStrictEqual({
+        behavior: 'auto',
```
the-t-in-rtf, line 64: This should also be reused, here it makes sense to define a constant for the expected value.

src/CharacterState.js
```diff
@@ -92,28 +95,112 @@ export default class CharacterState {
...
+        }
+
         const newPathSegment = {
-            x1: this.xPos,
-            y1: this.yPos,
-            x2: this.xPos + xOffset,
-            y2: this.yPos + yOffset
+            x1: startingX1,
+            y1: startingY1,
+            x2: newXPos,
```
the-t-in-rtf, line 139: Side note, I like this much better than performing the math inline.

src/CharacterState.js
```diff
@@ -122,7 +209,111 @@ export default class CharacterState {
...
+            this.xPos,
+            newYPos,
+            this.direction,
+            this.path,
+            this.sceneDimensions
+        );
+    }
+
+
+    getRowLabel(): string {
```
the-t-in-rtf, line 312: I'm a little uneasy having the type `string` for the return values here.  It seems like we could protect ourselves better by enumerating the valid values in `types.js`.

src/ProgramBlockEditor.js
```diff
@@ -35,6 +43,9 @@ type ProgramBlockEditorProps = {
     focusTrapManager: FocusTrapManager,
     addNodeExpandedMode: boolean,
     theme: string,
+    onChangeCharacterPosition: (direction: ?string) => void,
+    onChangeCharacterXPosition: (columnLabel: string) => void,
```
the-t-in-rtf, line 47: The type should also be used here.

src/ProgramBlockEditor.js
```diff
@@ -523,11 +589,17 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
 
     getThemedCharacter() {
         if (this.props.theme === 'space') {
-            return <SpaceShipIcon className='ProgramBlockEditor__character-column-character' />
+            return <SpaceShipIcon
+                transform={`rotate(${this.props.characterState.getDirectionDegrees() - 90} 0 0)`}
```
the-t-in-rtf, line 593: In my own work in this area, I thought I had seen a variation where the transformation was defined as a variable and reused.  That seems more appropriate.

src/ProgramBlockEditor.js
```diff
@@ -570,14 +650,92 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
...
+                            onClick={!this.props.editingDisabled ? this.handleClickCharacterPosition : undefined} />
+                    </div>
+                    <div className='ProgramBlockEditor__character-move-position-coordinate'>
+                        <input
+                            className={characterPositionTextInputClassName}
+                            aria-label={this.props.intl.formatMessage({id:'ProgramBlockEditor.editPosition.columnPosition'})}
+                            aria-disabled={this.props.editingDisabled}
+                            type='text'
+                            value={this.state.characterColumnLabel}
+                            onChange={!this.props.editingDisabled ? this.handleChangeXPos : undefined}
```
the-t-in-rtf, line 728: Given that the change doesn't take effect until you blur the control, I'm not sure that `onChange` is adding anything.  In fact you might be attempting to update the position twice.

As discussed elsewhere, we should have support for hitting enter as well.  I think `onBlur` and a small `onKeyDown` handler would be good, unless you have a concrete reason for using `onChange` related to a particular browser.

the-t-in-rtf:
I left a few comments on the code, but found a fairly big blocker afterwards.  The character is not visible at all on the scene in Firefox.  The path is visible, and based on that the position does appear to be correct, but the character is never visible.

src/svg/TurnPositionLeft.svg
```diff
@@ -0,0 +1,16 @@
...
+</g>
+<defs>
+<filter id="filter0_d" x="0.5" y="0" width="26" height="14.9793" filterUnits="userSpaceOnUse" color-interpolation-filters="sRGB">
```
the-t-in-rtf, line 6: Given our recent experience with the play and stop buttons, we should remove any filters we can.  Do these add anything?

src/svg/TurnPositionRight.svg
```diff
@@ -0,0 +1,16 @@
...
+</g>
+<defs>
+<filter id="filter0_d" x="0.5" y="0" width="26" height="15" filterUnits="userSpaceOnUse" color-interpolation-filters="sRGB">
```
the-t-in-rtf, line 6: Same question here, is this filter necessary?

src/svg/TurnPositionLeft.svg
chosww: Screen Shot 2021-03-17 at 1 15 22 PM

src/Character.js
chosww: This change was introduced in 286, so we should resolve any issue to this in 286; however, in Character component, the scrollIntoView property object is used in two places, and they have different scopes, so we can't have a variable for the object, unless we specify that in a class property.

src/CharacterState.js
chosww: Currently, expected row label values are from 1 to 16, and this may change between any values from 1 to 26, based on our decision on how many rows we want in the scene. Specifying all those values as a type might be undesirable.

src/ProgramBlockEditor.js
chosww: onChange handler is responsible for updating values represented on the text box UI, but that doesn't actually change values of CharacterState. We don't want to apply changes as users change text box values, because that will send potentially invalid values for xPos and yPos in CharacterState. OnBlur will communicate with CharacterState state in App and update xPos and yPos in CharacterState after input validation.

src/svg/TurnPositionLeft.svg
the-t-in-rtf: Makes sense, then.

src/CharacterState.js
the-t-in-rtf: As long as it's clearly identified as providing a string to be used as a label, it's fine.  We can talk about adding more checks if we need them to help avoid actual issues.

src/Character.js
the-t-in-rtf: Thanks for the explanation.  Given that the Character only uses the function in a single way, a class property would seem to make sense.

src/ProgramBlockEditor.js
chosww: Good idea, I used a const for the transform.

chosww:
Thanks for reviewing the PR @the-t-in-rtf, I have made updates to address some issues you have found.

- Fixed an issue where the character draws a dot when trying to move outside of the scene from the edge.
- Fixed an issue where the character draws unexpected path when trying to move outside of the scene.
- Implemented new keyDown property on the coordinate text boxes to update the character position on enter key.
- Updated transform of the character in the character column to use const transform variable.

Some changes you suggested regard to the Character component should be resolved in C2LC-286, which Simon has been reviewing. I also want to see what Simon thinks about defining event specific object as a class property, before making the changes.

PR #135, opened 2021-03-30
C2LC-316: Fix focus trap settings for actions simplification menu toggle.

the-t-in-rtf:
See [C2LC-316](https://issues.fluidproject.org/browse/C2LC-316) for details.


the-t-in-rtf:
After our pairing session yesterday, I went further down the road with using references, and hit a few roadblocks.  Using `onBlur` proved pointless as there was never a `relatedTarget` and the `target` was the even losing focus.  So we had the wrong end of things, i.e. we only were notified when one of the action menu's elements blurred, and not when an outside element blurred.

After wrestling with a few variations, I ended up using an approach similar to our modal, i.e. when the menu is open, a large transparent element is positioned to fill the screen, just behind the menu itself.  Clicking anywhere outside the menu closes the menu and also removes the transparent element to avoid side effects.

I also made a first pass at ensuring that the first checkbox in the panel is focused when the menu is open.  I could use help with the polish on this, the focus works as expected when opening the menu with the keyboard, but when clicking the toggle with a mouse, the focus seems to move but the highlighting is wrong in Chrome.

the-t-in-rtf:
> the focus works as expected when opening the menu with the keyboard, but when clicking the toggle with a mouse, the focus seems to move but the highlighting is wrong in Chrome.

As discussed in the chats, it seems like the focus position moves when you click the toggle using the mouse.  Even though there are no `:focus` styles applied:

1. If you hit space, the first checkbox is toggled.
2. If you hit tab, focus moves to the second checkbox.

Oddly, if you hit a key such as caps lock, shift, arrows, the focus highlighting will appear on the correct element.

sbates-idrc:
I'm seeing an issue on Firefox on Windows (which isn't happening on Chrome on Windows): when I open the menu using the spacebar, the spacebar key press toggles the first item in addition to opening the menu.

sbates-idrc:
@the-t-in-rtf Could you expand a little on the point above:

"If I reuse the global focus trap manager, it's unusable as it gets flagged as inactive by the program block editor. So we do need to use our own."

Thanks.

the-t-in-rtf:
> If I reuse the global focus trap manager, it's unusable as it gets flagged as inactive by the program block editor. So we do need to use our own.

If I have App.js pass in the focus trap manager, the program block editor deactivates it (sets `active` to false) whenever the focus manager doesn't need it, which causes the focus trap manager to ignore all keystrokes until it's activated again.  My gut feel is that the opposite situation (the actions menu deactivating the focus trap manager when the program block editor needs it) is far less likely, but as long as the focus trap manager has a single toggle to control whether or not it's listening, it doesn't seem safe to have two components rely on the same instance.

the-t-in-rtf:
From chats with @sbates-idrc:

> Maybe we don't need/want to set focus? If I am understanding your comments on the PR, I think you aren't using blur any longer to close the menu, is that right? When we were working on it yesterday, setting focus on the first item was motivated by using blur to close. And if we aren't using blur, then maybe we don't need to set focus. Just set up the focus trap and ensure that the first menu control is in the tab order after the button to open it. For example, for the popup for the program steps, we don't set focus in that case. Screen reader users I believe are generally the most sensitive to setting focus programmatically, and I think if we don't need to do it, that might be a better user experience.

In other messages, Simon also noted that if you activate the menu by pressing the space key on Firefox in Windows, the first item gets toggled, which I verified as well on the Mac.  Given this side effect, I think Simon's argument above for removing the "focus on first" logic is even stronger.

To answer the question, yes, now we use a click on an external element to close the menu, so we don't have a good reason to fiddle with the focus, and I agree that it's to be avoided if possible.  I will update the pull accordingly, and see if my idea about using the global focus trap manager works as well.

the-t-in-rtf:
> I will update the pull accordingly, and see if my idea about using the global focus trap manager works as well.

With the approach I proposed, the shared focus trap manager remains active, which is good, but there's a side effect, suddenly every tab key moves forward two items (and every shift+tab moves backwards by two items), which makes every other checkbox unreachable.

 I pushed up the latest changes without the shared focus trap manager changes, so that the merge would be otherwise merge-ready.  It only takes a minute to refactor, we can pair on it and look at it together if you want.

sbates-idrc:
@the-t-in-rtf I removed some more code that is no longer needed: https://github.com/sbates-idrc/c2lc-coding-environment/commit/e37d40de005c97fa6f2c1f444d683b7ca36e4b9a

Please take a look and let me know if it looks good. I will then merge. Thanks.

the-t-in-rtf:
@sbates-idrc, thanks for catching that, looks fine to me.

the-t-in-rtf:
@sbates-idrc, I merged in your changes and updated with `develop-0.7` to pick up other fixes.

PR #140, opened 2021-03-31
C2LC-317: Adjusted sound panning to match new scene dimensions.

the-t-in-rtf:
See [C2LC-317](https://issues.fluidproject.org/browse/C2LC-317) for details.

### DIFF
--- a/src/AudioManagerImpl.js
+++ b/src/AudioManagerImpl.js
@@ -162,7 +162,12 @@ export default class AudioManagerImpl implements AudioManager {
             // As we use a single Sampler grade, our best option for panning is
             // to pan all sounds.  We can discuss adjusting this once we have
             // multiple sound-producing elements in the environment.
-            const panningLevel = ((characterState.xPos / sceneDimensions.getWidth()) * 2) - 1;
+
+            // TODO: Update this to use maxX - minX once the scene properly exposes the minimum width.
+            const midPoint = ((sceneDimensions.getWidth() + 1 )/ 2)
+
+            // Limit the deviation from the centre so that there is always some sound in each speaker.
+            const panningLevel = 0.65 * ((characterState.xPos - midPoint) / midPoint);
 
             // TODO: Consider making the timing configurable or tying it to the movement timing.
             this.panner.pan.rampTo(panningLevel, 0.5)
--- a/src/AudioManagerImpl.js
+++ b/src/AudioManagerImpl.js
@@ -4,6 +4,7 @@ import { Midi, Panner, Sampler } from 'tone';
 import CharacterState from './CharacterState';
 import type {IntlShape} from 'react-intl';
 import {AudioManager} from './types';
+import SceneDimensions from './SceneDimensions';
 
 function octaveModulo (rawPitch: number) : number {
     const adjustedPitch = rawPitch % 12;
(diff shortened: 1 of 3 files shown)


src/AudioManagerImpl.js
```diff
@@ -161,7 +162,7 @@ export default class AudioManagerImpl implements AudioManager {
             // As we use a single Sampler grade, our best option for panning is
             // to pan all sounds.  We can discuss adjusting this once we have
             // multiple sound-producing elements in the environment.
-            const panningLevel = Math.min(1, Math.max(-1, (0.1 * characterState.xPos)));
+            const panningLevel = ((characterState.xPos / sceneDimensions.getWidth()) * 2) - 1;
```
sbates-idrc, line 165: The `panningLevel` calculation isn't quite symmetrical. The `xPos` varies from 1-26 inclusive.

- At `xPos` of 1, `panningLevel` is -0.92
- At `xPos` of 26, `panningLevel` is 1

We'd want a calculation like:

    const panningLevel = ((characterState.xPos - minX) / (maxX - minX) * 2) - 1;

Where `minX` and `maxX` are from the `SceneDimensions`. Unfortunately, right now `SceneDimensions` will not give us the right min X and max X. It may be too risky to rework `SceneDimensions` at this point in the 0.7 timeline. I have another thought on panning, which I will leave as a separate comment.

sbates-idrc: In the new coordinate system, the `SceneDimensions` constructor should really now be:

    constructor(minX: number, maxX: number, minY: number, maxY: number)

So we can construct with:

    new SceneDimensions(1, 26, 1, 16);

But as I mentioned above, it may be too risky to rework at this stage.

sbates-idrc:
I asked @chosww to help me test this PR as he initially reported the issue. And he was initially still not hearing sound at the last column Z. As we talked, we figured out that it was because he was testing with only 1 earbud in: the left. Which made me think that hard panning from full left to full right might not be quite the best thing. As it will create a worse experience for people only hearing one of the speakers/ears. Some ideas:

- Pan but reduce the range so that some sound is audible at all times on the L and R channels
- Or, remove panning for now and potentially add an option in the future to turn panning on/off

sbates-idrc:
I asked Sepideh for her thoughts on panning in the C2LC General room on Thursday April 1 and she would like to keep the panning for 0.7 as she would like to try it out with the workshop co-designers. So, we either keep the full panning from full left to full right, or we maybe try a slightly reduced range, so that some audio is always present in each ear.

the-t-in-rtf:
Thanks for all the feedback, I will work on improving the panning so that it never drops below 25% in either ear, and so that it's properly symmetrical.

the-t-in-rtf:
I added a limiter on the panning range so that there is always at least some sound in each ear, but so that the panning still clearly indicates which "side" of the median the character is on. I also ensured that the panning is symmetrical.

src/AudioManagerImpl.js
the-t-in-rtf: I hard-coded the current minimum in an adjusted calculation of the midpoint, so that the panning is symmetrical.

I also left a TODO in the relevant area about using minX and maxX.

PR #144, opened 2021-04-01
C2LC-324: Commented out world picker header icon.

the-t-in-rtf:
See [C2LC-324](https://issues.fluidproject.org/browse/C2LC-324) for details.

### DIFF
--- a/src/WorldSelector.js
+++ b/src/WorldSelector.js
@@ -6,7 +6,8 @@ import type { IntlShape } from 'react-intl';
 import { injectIntl } from 'react-intl';
 
 import AriaDisablingButton from './AriaDisablingButton';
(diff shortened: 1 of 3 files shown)


the-t-in-rtf:
I commented out the world picker header icon with clear TODO notes about it, and adjusted the spacing of the first world icon as well.  @sbates-idrc, this should be ready for review.

the-t-in-rtf:
The first build failed because the tests try to click buttons based on their position within the world selector, which changed when the header icon was commented out.  There was also a test for the header icon's ARIA label that failed.  I have commented that out, and adjusted the other tests with a note about what needs to be done when the header icon is reenabled.

PR #154, opened 2021-04-06
C2LC-337: Announce updated Scene’s aria-label at the end of program runs

chosww:


the-t-in-rtf:
I tested this in detail with VoiceOver, and in most circumstances, it behaved as expected:

1. After running a single step program, the position (and orientation) is announced.
2. After running a multi-step program, the position is announced.
3. When pausing, the position is announced.
4. When resuming, the program completes and the position is announced.
5. After refreshing the scene when the character has moved, the position is announced.
6. Hitting refresh when the character is already in the starting position does not result in an announcement.
7. If the program has already been run, using any of the position controls announces the position (see below).

However, I seem to be able to get into a state where the changes made by the position controls are not announced.  Steps to reproduce:

1. Hack the URL and set `p` to nothing.
2. Move the character around and turn it using the position controls.   Nothing is announced.
3. Enter a program and play it.
4. Move the character around and turn it using the position controls.  The position and orientation are announced each time they change.

I will look at the code as well to see if I can find an explanation.

src/Scene.js
```diff
@@ -268,6 +281,20 @@ class Scene extends React.Component<SceneProps, {}> {
             </React.Fragment>
         );
     }
+
+    componentDidUpdate(prevProps: SceneProps, prevState: SceneState) {
+        if (prevProps.runningState !== this.props.runningState) {
```
the-t-in-rtf, line 286: This is not sufficient to ensure that position changes are announced before the program has been run.  You'll also need to check to see if the character position/orientation within the scene have changed.  As it stands, changes that result from using the position controls are not announced until you run the program at least once.

the-t-in-rtf:
With the latest changes, the behaviour is now what I'd expect on "startup", i.e changes made using the positional controls are always announced.

public/index.html
```diff
@@ -5,6 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <link rel="apple-touch-icon" href="logo192.png" />
+    <link rel="stylesheet" href="index.css" />
```
the-t-in-rtf, line 8: Don't we already have an `index.scss`?

chosww: This is inside of public folder, as the style is applying to the index.html file in public.

the-t-in-rtf: I guess my question is, why do we need a new file?

the-t-in-rtf: I mean, if I load the actual site, I see both this include and the webpack "chunk", which means another XHR request, cache check, et cetera.  Does it not work if the rules are in `index.scss` and the styles are packed up with everything else?

sbates-idrc: I think we should be able to remove the custom `index.css` and use the Boostrap `sr-only` class directly.

src/ProgramBlockEditor.test.js
```diff
@@ -134,6 +139,7 @@ function createMountProgramBlockEditor(props) {
             )
         ),
         {
+            attachTo: ariaLiveDiv,
```
sbates-idrc, line 142: I believe that this `attachTo` should not be necessary.

PR #163, opened 2021-04-09
C2LC-347: Fix character scrolling in Safari.

the-t-in-rtf:
See [C2LC-347](https://issues.fluidproject.org/browse/C2LC-347) for details.


the-t-in-rtf:
To summarise the research I did and approach I used:

1. Various anecdotal evidence suggests we're not along here [for example, this post](https://stackoverflow.com/questions/60418768/in-safari-how-to-scroll-svg-child-element-into-view).
2. I found [another post](https://www.gitmemory.com/issue/iamdustan/smoothscroll/153/611965308) that suggested rewiring the SVG prototype for Safari or using a polyfill.
3. The polyfills I reviewed seemed really heavyweight to drop in without careful review.

So, I ended up having the scene keep track of whether the character was in position and scrolling itself accordingly.  To the extent it works, it works in Safari as well.  There are quite a few issues, I'm putting in the pull so that a preview site can be built and we can discuss.  Among other things:

1. Although moving out of bounds to the left often triggers the expected scrolling behaviour, it's possible to move the character off the scene to the left.
2. Moving from left to right along the top boundary seems to be triggering up/down scrolling in a way that's undesirable.
3. When the scrolling works, it jumps very fall all at once.  It would be nicer if it adjusted "just enough".

I'm sure there are other issues as well, I suspect we may have to live with workarounds and dig into this in 0.8.

the-t-in-rtf:
The tests failed because I haven't removed the old focus tests from Character.test.js.

the-t-in-rtf:
I tested this with the iPad Air in the Simulator on OS X.  Without this fix, scrolling doesn't work.  With this fix, scrolling does work.

src/Character.js
```diff
@@ -55,25 +50,10 @@ export default class Character extends React.Component<CharacterProps, {}> {
     render() {
         return (
             <g
-                ref={this.characterRef}
```
the-t-in-rtf, line 58: As discussed in our pairing session, we should work to restore the use of refs instead of IDs if possible, as we want to have multiple characters eventually, and won't have a single ID in that case.

the-t-in-rtf:
In testing yesterday, I couldn't get access to the DOM element using a $ref (only the React component was visible).

I ended up adding a distinct ID to each icon, and then making the ID that corresponds to the icon available in a small getter.  With this approach, I was able to retrieve the ID from the component, and then use document.getElementById to get the DOM element.

the-t-in-rtf:
As mentioned in our last planning meeting, we need to agree how to proceed here.  Daniel prepared an alternate approach in [his C2LC-347 branch](https://github.com/chosww/c2lc-coding-environment/tree/C2LC-347), which also moves the logic back to the character.   If we agree this is where we want to solve this (and how), I can merge with his work and fix the Flow errors that broke his last build.

sbates-idrc:
This PR is looking great. Thank you.

I have one small change I'd like before merging: the scroll calculation uses an offset of "50". I'd like to replace this with something that is calculated from the Scene contents (to protect us from changes in the scale of the Scene). Am I right in thinking that this is to provide some space around the character as the scene is scrolled? Would this then be dependent on the size of the character? Could we use `newCharacterBounds.width` and `newCharacterBounds.height` as the scroll offset?

the-t-in-rtf:
> Could we use `newCharacterBounds.width` and `newCharacterBounds.height` as the scroll offset?

Good idea, and done.

sbates-idrc:
I found an issue while testing this PR today: the scrolling behaves differently when the character is rotated at NE, SE, SW, or NW versus N, E, S, or W. When the robot character is facing NE, SE, SW, or NW, the character bounding box is 70 x 70. Whereas, when it is facing N, E, S, or W, the character bounding box is 49.5 x 49.5.

I have an idea for a solution: instead of using a ref on the Character SVG, we could instead put a ref on the rectangle that we are using for the character background, and use that for our scrolling calculation:

```
<rect
    className="Character__icon-background"
    x={-0.5}
    y={-0.5}
    height={1}
    width={1}
    transform={characterBackgroundTransform}
/>
```

This rectangle will have a constant size; it will be independent of character rotation, or details of the character SVG.

the-t-in-rtf:
> I have an idea for a solution: instead of using a ref on the Character SVG, we could instead put a ref on the rectangle that we are using for the character background

Very good idea, thanks.  Because that's HTML rather than a component, we also have to fight with React a lot less to get to the information we need.  I've removed all the ref infrastructure from the Character and changed the names to reflect what we're using for our calculations.  Since the background always matches the cell dimensions, this seems pretty clean in my testing.

PR #168, opened 2021-04-14
C2LC-351: Fix an issue with the character in the scene could draw other angles than 45 degree

chosww:


the-t-in-rtf:
See [C2LC-351](https://issues.fluidproject.org/browse/C2LC-351) for details.

Thanks for including [the sample URL to reproduce the problem](https://create.weavly.org/?v=0.7&t=mixed&w=space&p=3A6a34B3a6a1B3a6&c=aab&a=123456ABDabd), that was very helpful.  Here's the program run once from A1 facing right in the live site:

Screenshot 2021-04-21 at 10 42 34

Here's [the same program in your preview site](https://deploy-preview-168--c2lc-build.netlify.app/?v=0.8&t=mixed&w=space&p=3A6a34B3a6a1B3a6&c=aab&a=123456ABDabd), run once from A1 facing right:

Screenshot 2021-04-21 at 10 42 55

I tried a few more variations, J1 facing right, A1 at various orientations.  I couldn't get it any non-standard angles to appear.

src/CharacterState.js
```diff
@@ -98,41 +98,48 @@ export default class CharacterState {
...
-                    turnSegment = this.drawEdgeDiagonalPath([], newYPos - this.yPos, this.direction);
-                }
-                break;
-            default:
-                newYPos = this.yPos + yOffset;
-                break;
+        const boundsStateX = this.sceneDimensions.getBoundsStateX(this.xPos + xOffset);
+        const boundsStateY = this.sceneDimensions.getBoundsStateY(this.yPos + yOffset);
+
+        if (boundsStateX === 'outOfBoundsBelow' && boundsStateY === 'outOfBoundsBelow') {
```
the-t-in-rtf, line 104: I think this could be expressed more cleanly, perhaps by splitting up the logic by X and Y.  For example, in all cases where `boundsStateX` is `outOfBoundsBelow`, `newXPos` gets set to `1`, and yet you set it three times.

chosww: We could group all the cases where boundsStateX is the same, but that will introduce condition checks for different boundStateY inside of the condition, which may make reading the code more difficult. I think writing all possible combinations with some redundancy has better readability.

PR #169, opened 2021-04-14
C2LC-355: Remove user select from the App

chosww:


the-t-in-rtf:
See [C2LC-355](https://issues.fluidproject.org/browse/C2LC-355) for details.  I fleshed out the ticket, as there are actually a range of elements that can trigger this behaviour:

1. Program steps in the program sequence container.
2. The scene.
3. The stop button (when disabled)
4. Expanded add node buttons
5. Any scene column or row header.
6. The program header
7. The "Weavly" text.

Thankfully, the fix seems to have addressed each of these, and I couldn't find any side effects.

PR #170, opened 2021-04-14
C2LC-356: Disable double tap zoom in iPad

chosww:


the-t-in-rtf:
See [C2LC-356](https://issues.fluidproject.org/browse/C2LC-356) for details.

the-t-in-rtf:
In my testing, this fixes the zooming in the iPad simulator, although I had to come up with my own steps to reproduce.  It would be good to a) verify on an iPad and b) add more details on other ways to reproduce to the JIRA ticket.

chosww:
I have verified that zooming is not happening on an iPad for the scenario you have written on the ticket.

sbates-idrc:
I have verified that this PR does stop the zooming on double tap. However, it has some effects on other touch gestures. It interferes with the pinch to zoom and panning, which I think we would want to enable a user to still do. Without this change, pinch to zoom and panning worked on the application. With this change, only some parts of the interface react to pinch to zoom and panning.

For example, what does still work:

- Pinch to zoom in with both fingers within the program editor
- One can zoom out and pan with fingers within the program editor

Something that used to work but no longer does:

- Pinch to zoom in with one finger within the program editor and one finger on the program heading bar (the element containing the "Program" heading, node toggle switch, and Delete All button)

chosww:
Thanks @sbates-idrc for the review. I found out that there is manipulation option for touch-action, that disables double tap zooming only. "Enable panning and pinch zoom gestures, but disable additional non-standard gestures such as double-tap to zoom. Disabling double-tap to zoom removes the need for browsers to delay the generation of click events when the user taps the screen." 

After the change, I tried and confirmed that the scenario you described on your comment works fine now.

PR #171, opened 2021-04-19
C2LC-357: Fix an issue with Command blocks have no vertical space between them in portrait mode on the iPad

chosww:


the-t-in-rtf:
The fix definitely improves the portrait mode look in the iPad simulator, but I'm wondering why we need to go to one column of blocks at all.  The width of the container never changes, as shown here:

Command Blocks Column

Unless we are also somehow decreasing the width of the container, to me it makes as much sense to just have it always be two columns.

chosww:
Yeah, you are totally right about we don't even need to change the layout when screen size gets narrower. But we have a separate issue for rework of narrower layout, and this is just a momentary fix until we work on that.

the-t-in-rtf:
Thanks, I figured it might have to do with that.

PR #157, opened 2021-04-07
C2LC-332: Fix turn control 'disabled' styles.

the-t-in-rtf:
See [C2LC-332](https://issues.fluidproject.org/browse/C2LC-332) for details

PR #172, opened 2021-04-19
C2LC-352: Apply proper disabled styles

chosww:


the-t-in-rtf:
First, I see that the changes in #157 aren't duplicated here.  (It's fine, just making a note of the scope.)

Screenshot 2021-04-21 at 10 25 03

I see that the "delete all" and "refresh" buttons now (correctly) lack hover styles when disabled, that's great.

The one thing I see that needs addressing is that the highlighting for the selected "world" disappears when the program is running.

chosww:
Thanks Tony, I talked with Sepideh about the disabled style for the world selector, and she thinks whether a world is selected or not, they should have same disabled style.

sbates-idrc:
I found some issues that I think we should address before merging this PR:

- The Refresh button and the World Selector character buttons change colour when clicked when they are disabled. Clicking the Delete All button when it is disabled does not change the colour (in all themes except High Contrast, where it does change colour, see below).
- In High Contrast, the Delete All button does change colour when clicked when disabled.
- In High Contrast, the Refresh button has a hover style that is visible when it has focus. To reproduce:
    - Turn on High Contrast
    - Run a program so that the Refresh button is disabled
    - Set focus to the Refresh button
    - Hover over the Refresh button

PR #173, opened 2021-04-20
C2LC-346: Add timeout to pushState to ensure that the function is not…

chosww:
… called more than 100 times in 30 seconds

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -92,7 +92,7 @@ export class App extends React.Component<AppProps, AppState> {
     characterStateSerializer: CharacterStateSerializer;
     allowedActionsSerializer: AllowedActionsSerializer;
     speedLookUp: Array<number>;
-    pushStateTimer: ?TimeoutID;
+    pushStateTimeoutID: ?TimeoutID;
 
     constructor(props: any) {
         super(props);
@@ -118,7 +118,7 @@ export class App extends React.Component<AppProps, AppState> {
 
         this.allowedActionsSerializer = new AllowedActionsSerializer();
 
-        this.pushStateTimer = null;
+        this.pushStateTimeoutID = null;
 
         this.interpreter.addCommandHandler(
             'forward1',
@@ -987,8 +987,10 @@ export class App extends React.Component<AppProps, AppState> {
             const serializedProgram = this.programSerializer.serialize(this.state.programSequence.getProgram());
             const serializedCharacterState = this.characterStateSerializer.serialize(this.state.characterState);
             const serializedAllowedActions = this.allowedActionsSerializer.serialize(this.state.allowedActions);
-            clearTimeout(this.pushStateTimer);
-            this.pushStateTimer = setTimeout(() => {
+            // Safari throws an error: "SecurityError: Attempt to use history.pushState() more than 100 times per 30 seconds".
+            const pushStateDelayMs = 350;
+            clearTimeout(this.pushStateTimeoutID);
+            this.pushStateTimeoutID = setTimeout(() => {
                 window.history.pushState(
                     {
                         p: serializedProgram,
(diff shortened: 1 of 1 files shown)


the-t-in-rtf:
OK, I testing this and verified that:

1. It's no longer possible to trigger the error by running a long program (big square) at the highest speed.
2. It's no longer possible to trigger the error by hitting the rotate button continuously (I use that because it always changes, where the positional controls hit the edge).

I tried refreshing very quickly after rotating the character a few dozen times, I see that it's possible to lose a little bit of state, one or two steps, this is probably acceptable.    This probably needs to be filed as a new ticket, but I did notice that our URL persistence strategy doesn't let you undo changes using the back button.

I will move on and look at the code shortly.

src/App.js
```diff
@@ -984,18 +987,23 @@ export class App extends React.Component<AppProps, AppState> {
...
-                    c: serializedCharacterState,
-                    t: this.state.settings.theme,
-                    a: serializedAllowedActions,
-                    w: this.state.settings.world
-                },
-                '',
-                Utils.generateEncodedProgramURL(this.version, this.state.settings.theme, this.state.settings.world, serializedProgram, serializedCharacterState, serializedAllowedActions),
-                '',
-            );
+            if (this.pushStateTimer == null) {
```
the-t-in-rtf, line 990: This addresses the problem, but in thinking about it more I have a concern that we'd end up not persisting the last change in the series, which is the most critical to preserve.

I'd suggest inverting the logic, if `this.pushStateTimer` is not `null` (i.e. the last update hasn't completed), use `clearTimeout(this.pushStateTimer)` to stop the previous timer, and then always set `this.pushStateTimer` as you're doing here.

That way the last change (the final state) has a better chance of winning if there's a "race".

Another strategy would be to make it queue-based, i.e. push all the state changes to a stack and pop one off every 300ms until there are none left.  Probably that's a bit much for this go.

the-t-in-rtf: > Another strategy would be to make it queue-based, i.e. push all the state changes to a stack and pop one off every 300ms until there are none left. Probably that's a bit much for this go.

I thought more about this, and @simonbates also noticed that the queue strategy would push the out the time at which all events are persisted and make it more likely that we'd lose things.

chosww: Thanks @the-t-in-rtf and @simon-idrc, for the feedback, I will rework on the logic to ensure last state gets stored.

src/App.js
```diff
@@ -92,6 +92,7 @@ export class App extends React.Component<AppProps, AppState> {
     characterStateSerializer: CharacterStateSerializer;
     allowedActionsSerializer: AllowedActionsSerializer;
     speedLookUp: Array<number>;
+    pushStateTimer: ?TimeoutID;
```
sbates-idrc, line 95: I'm thinking renaming this to `pushStateTimeoutID` would make the intention clearer.

src/App.js
```diff
@@ -984,18 +987,21 @@ export class App extends React.Component<AppProps, AppState> {
...
-                    c: serializedCharacterState,
-                    t: this.state.settings.theme,
-                    a: serializedAllowedActions,
-                    w: this.state.settings.world
-                },
-                '',
-                Utils.generateEncodedProgramURL(this.version, this.state.settings.theme, this.state.settings.world, serializedProgram, serializedCharacterState, serializedAllowedActions),
-                '',
-            );
+            clearTimeout(this.pushStateTimer);
```
sbates-idrc, line 990: Please add a comment explaining why this logic is here. Safari throws an error: "SecurityError: Attempt to use history.pushState() more than 100 times per 30 seconds".

src/App.js
```diff
@@ -984,18 +987,21 @@ export class App extends React.Component<AppProps, AppState> {
...
+                        c: serializedCharacterState,
+                        t: this.state.settings.theme,
+                        a: serializedAllowedActions,
+                        w: this.state.settings.world
+                    },
+                    '',
+                    Utils.generateEncodedProgramURL(this.version, this.state.settings.theme, this.state.settings.world, serializedProgram, serializedCharacterState, serializedAllowedActions),
+                    '',
+                );
+            }, 300);
```
sbates-idrc, line 1004: I think we should increase the value of the delay by a little bit, as 300ms is right at the limit of 100 times in 30 seconds. Say to 350ms. I think that it would also be worthwhile to make a `const` variable for this value so that we can give it a name: `pushStateDelayMs`.

PR #174, opened 2021-04-21
C2LC-365: Clean up test output.

the-t-in-rtf:
See [C2LC-365](https://issues.fluidproject.org/browse/C2LC-365) for details.

Sorry about the goofy ticket number on the first commit, I had to cherry pick it from a commit in the wrong branch.


chosww:
Thanks @the-t-in-rtf, I did some research and found out that we can use "is" attribute to tell that g tag is a custom element and that way we wont get the unrecognized tag warning. We don't have to actually define what that custom element is, but just use a reasonable name for "is" attribute, like 'svg-group'.

chosww:
We should wait for #175 to be merged into development-0.8 branch and merge changes for unhandled promise in the current 0.8 branch together.

the-t-in-rtf:
> Thanks @the-t-in-rtf, I did some research and found out that we can use "is" attribute to tell that g tag is a custom element and that way we wont get the unrecognized tag warning. We don't have to actually define what that custom element is, but just use a reasonable name for "is" attribute, like 'svg-group'.

That's very helpful, thanks @chosww.

sbates-idrc:
Please merge in the latest `develop-0.8` and resolve conflicts.

the-t-in-rtf:
Done.

src/Character.js
```diff
@@ -48,6 +48,7 @@ export default class Character extends React.Component<CharacterProps, {}> {
         return (
             <g
                 className='Character'
+                is='svg-group'
```
sbates-idrc, line 51: The "is" attribute is for specifying custom HTML elements:

https://html.spec.whatwg.org/multipage/custom-elements.html#customized-built-in-element

I don't think that this is the right approach for silencing the warning from the test as it may have undesirable consequences for the running application, by triggering the custom element processing within the browser.

In this case, I'd rather just live with the warning, if we can find no other solution.

public/index.html
```diff
@@ -23,6 +23,11 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+    <!-- Workaround for Tone.js logging. -->
+    <script type="text/javascript">
+      window.TONE_SILENCE_LOGGING = true;
```
sbates-idrc, line 28: This setting of `TONE_SILENCE_LOGGING` will take effect for the running application but does not change the tests. I still see logging in the test output, such as:

```
PASS  src/App.test.js
  ● Console

    console.log
       * Tone.js v14.7.77 *

      at Module.o (node_modules/tone/build/webpack:/Tone/Tone/core/Global.ts:72:10)
```

sbates-idrc:
I'm still seeing the following:

```
 PASS  src/ProgramBlockEditor.test.js
  ● Console

    console.log
       * Tone.js v14.7.77 *

      at Module.o (node_modules/tone/build/webpack:/Tone/Tone/core/Global.ts:72:10)

    console.error
      Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
          in input (at ProgramBlockEditor.js:729)
          in div (at ProgramBlockEditor.js:728)
          in div (at ProgramBlockEditor.js:659)
          in div (at ProgramBlockEditor.js:631)
          in ProgramBlockEditor (created by Context.Consumer)
          in injectIntl(ProgramBlockEditor) (created by WrapperComponent)
          in RootFinder (created by WrapperComponent)
          in IntlProvider (created by WrapperComponent)
          in WrapperComponent

      at printWarning (node_modules/prop-types/checkPropTypes.js:20:15)
      at checkPropTypes (node_modules/prop-types/checkPropTypes.js:82:11)
      at Object.<anonymous>.ReactControlledValuePropTypes.checkPropTypes (node_modules/react-dom/cjs/react-dom.development.js:1679:5)
      at initWrapperState (node_modules/react-dom/cjs/react-dom.development.js:1832:35)
      at setInitialProperties (node_modules/react-dom/cjs/react-dom.development.js:5990:7)
      at finalizeInitialChildren (node_modules/react-dom/cjs/react-dom.development.js:7499:3)
      at completeWork (node_modules/react-dom/cjs/react-dom.development.js:18978:17)
      at completeUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:22189:16)
```

Is this one that we can address?

src/Character.js
the-t-in-rtf: OK, I'll take it out.

the-t-in-rtf:
> Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.

We can if we use `() => {}` instead of `undefined`.

public/index.html
the-t-in-rtf: The only way I could find to get rid of this was to make `AudioManagerImpl` a dynamic import.  You can see what I've done in the latest commits.  If it's too much, we'll have to live with it, as I couldn't find any other way to set that variable before Tone got a chance to fire its startup message on the console.  I tried:

1. Updating the global window object provided by Enzyme.
2. Mocking the global window object and having it reply with `true` when queried for that property.
3. Temporarily reassigning `console.log` to a dummy function.

the-t-in-rtf:
@sbates-idrc, I addressed your feedback as best I could.  The dynamic import of `AudioManagerImpl` fixes all but one of the warnings (in the tests for the `AudioManagerImpl` itself).  I'm not sure you'll be happy with the solution, but thought I'd at least let you see it.

The `value`/`onChange` fix is hopefully less controversial, but I understand if that's also going too far.

Anyway, this is ready for further review at least.

PR #176, opened 2021-04-22
C2LC-353: Added character position control hover styles for all themes.

the-t-in-rtf:
See [C2LC-353](https://issues.fluidproject.org/browse/C2LC-353) for details.

Note that the "high contrast" theme was meant to have black fill, white stroke, but that only works for the turn controls. The arrows are "stroked" shapes with no fill, if you fill them, you get a solid (and very sharply angled) triangle.  I ended up cloning the existing path superimposed on the first and reducing its stroke-width.  This "inner" path is almost always the same colour as the "outer" path.  When hovering in the high contrast theme, the "inner" path is black and the "outer" path is white, which matches the effect used with the turn buttons as closely as I could manage.

### DIFF
--- a/src/Themes.scss
+++ b/src/Themes.scss
@@ -1474,12 +1474,14 @@ body.contrast-theme {
         color: black;
     }
 
-    .ProgramBlockEditor__character-position-button:not(.ProgramBlockEditor__character-position-button--disabled) {
-        stroke: white;
-
+    .ProgramBlockEditor__character-position-button {
         &:focus {
             outline: $c2lc-focus-indicator-width solid white;
-        }
+        }        
+    }
+
+    .ProgramBlockEditor__character-position-button:not(.ProgramBlockEditor__character-position-button--disabled) {
+        stroke: white;
 
         &:hover {
             path:not(.inner):not(.filled) {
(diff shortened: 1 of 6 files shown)


chosww:
The changes look good, but there may be some conflicts with position buttons disabled style work with the changed SVG. I think it makes sense to merge in this one before #172, and apply proper hover styles when buttons are disabled.

chosww:
Hey @the-t-in-rtf, development branch is ready to be merged with this work. After you merge them and resolve the conflicts, could you please update disabled reposition buttons have no hover styles? Thanks!

the-t-in-rtf:
I merged with develop-0.8 and fixed a few remaining issues with the high contrast and greyscale themes.  @chosww , this should be ready for further review.

src/Themes.scss
```diff
@@ -1462,19 +1476,47 @@ body.contrast-theme {
 
     .ProgramBlockEditor__character-position-button:not(.ProgramBlockEditor__character-position-button--disabled) {
```
chosww, line 1477: This rule is causing disabled reposition buttons to have no focus style for the contrast theme. We should specify the :not rule either inside of this rule, on a separate line.

the-t-in-rtf: Good catch!  Fixed.

chosww:
Thanks @the-t-in-rtf, everything looks good now.

PR #177, opened 2021-04-22
C2LC-359: Ensure that position is announced when returning to the same square and orientation.

the-t-in-rtf:
See[C2LC-359](https://issues.fluidproject.org/browse/C2LC-359) for details.

I tried what @chosww suggested, clearing the position text, but that did not result in a new announcement when it changed back.  I ended up announcing the fact that the character was moving, which does ensure that the position is announced again when returning to the same spot and orientation.


the-t-in-rtf:
To assist in verifying this, [here is a program that moves in a square returning to the original cell](https://deploy-preview-177--c2lc-build.netlify.app/?v=0.8&t=mixed&w=space&p=1b1b1b1b&c=aab&a=123456ABDabd).

chosww:
Hey @the-t-in-rtf and @sbates-idrc I did some research, and found out there are two potential fixes for this task. 

1. We could use politeness of assertive with the fix @the-t-in-rtf made, which includes announcement of 'Character is moving' -- assertive politeness would override any screen reader announcements, so pause won't be announced at all once the play button is pressed and that will fix weird announcement order in NVDA play -> robot character is moving -> pause -> robot character at column x row y.

2. We could go with our initial plan, which is just to remove the innerText when the play button is pressed, and then rewrite again after character's position is updated. I tried just removing the text and it wasn't working as Tony talked about, but I found out that we could have additional properties to aria-live region to make this work; there are two properties that aria-live supports, first one is aria-atomic, and second one is aria-relevant (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions). Setting aria-atomic to true works as what we want in NVDA, but doesn't work on VoiceOver, but setting aria-relevant to 'all' works on VoiceOver, but not on NVDA. So when I added both properties on the aria-live region, it worked fine on both NVDA and VoiceOver.

the-t-in-rtf:
I tried the second approach, which sometimes works, but not always. Sometimes no announcement is played.

I also noticed that there were double announcements when the "world" has just been updated:

1. Play the program once with the existing character, observe that the correct character, position and orientation are announced.
2. Change the character/world, not that the toggle change is announced but not the character + position.
3. Move focus back to the play button, let VoiceOver finish talking about the button.
4. Play the program.
5. The old character + position and the new character + position are announced.

Since the old and new positions are the same, it was difficult to tell exactly what's going on.  I tried breaking the "loop" and then it became apparent that both the old and new positions were being announced, and in that order.

the-t-in-rtf:
> 1. We could use politeness of assertive with the fix @the-t-in-rtf made, which includes announcement of 'Character is moving' -- assertive politeness would override any screen reader announcements, so pause won't be announced at all once the play button is pressed and that will fix weird announcement order in NVDA, play -> robot character is moving -> pause -> robot character at column x row y.

As discussed, I will proceed with (re)implementing option 1 and removing the variations we tried in working on option 2.

the-t-in-rtf:
I have updated to use the strategy we discussed. @chosww, can you please try this out again?

chosww:
Thanks @the-t-in-rtf, I checked that the new strategy is working fine in different browsers using VoiceOver; however, I see that NVDA is announcing "pause" before updated character position gets announced, even with assertive politeness.

I did some more research and found "aria-live="assertive" Any updates made to this region are important enough to be announced to the user as soon as possible, but it is not necessary to immediately interrupt the user." on https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/#liveprops, but that is outdated and new guide says "When regions are specified as assertive, assistive technologies will immediately notify the user, and could potentially clear the speech queue of previous updates." 

So I think omitting pause announcement might not be possible on NVDA.

sbates-idrc:
This PR will need to be updated now that C2LC-361, the change to move the code that updates the ARIA live region, has been merged into `develop-0.8`: https://github.com/codelearncreate/c2lc-coding-environment/commit/74673acf6effa375ab7c7c1ff663640c3144d505

The code to update the ARIA live region now lives in its own component: `CharacterAriaLive`

the-t-in-rtf:
Done.

src/messages.json
```diff
@@ -104,11 +104,13 @@
...
-        "ProgramBlockEditor.spaceShipCharacter": "Space Ship character at column {xPos}, row {yPos} facing {direction}",
-        "ProgramBlockEditor.rabbitCharacter": "Rabbit character at column {xPos}, row {yPos} facing {direction}",
-        "ProgramBlockEditor.robotCharacter": "Robot character at column {xPos}, row {yPos} facing {direction}",
-        "ProgramBlockEditor.editPosition.trunLeft": "Turn the character left",
-        "ProgramBlockEditor.editPosition.trunRight": "Turn the character right",
+        "ProgramBlockEditor.movementAriaLabel": "{character} is moving",
+        "ProgramBlockEditor.positionAriaLabel": "{character} at column {xPos}, row {yPos} facing {direction}",
+        "ProgramBlockEditor.spaceShipCharacter": "Space Ship character",
+        "ProgramBlockEditor.rabbitCharacter": "Rabbit character",
+        "ProgramBlockEditor.robotCharacter": "Robot character",
```
sbates-idrc, line 111: This is something that I didn't catch when we moved the ARIA live region code to the `CharacterAriaLive` component, but these messages should begin with "CharacterAriaLive" rather than "ProgramBlockEditor". Could we make that fix in this PR please?

src/CharacterAriaLive.js
```diff
@@ -15,53 +15,60 @@ type CharacterAriaLiveProps = {
...
-                    yPos,
-                    direction
-                }
+            return this.props.intl.formatMessage(
+                {id:'ProgramBlockEditor.robotCharacter'}
             );
         }
     }
 
+    clearCharacterPositionAriaLive() {
```
sbates-idrc, line 34: This method isn't clearing the character position, but rather setting the text to "{character} is moving". Please rename this method to better match its behaviour.

sbates-idrc:
Please add a test to `CharacterAriaLive.test.js` to cover the new message set for the `running` state.

src/CharacterAriaLive.js
the-t-in-rtf: I tried to avoid using words like "announce", I went with `setCharacterMovingAriaLive`.

src/messages.json
the-t-in-rtf: Done.

the-t-in-rtf:
> Please add a test to `CharacterAriaLive.test.js` to cover the new message set for the `running` state.

Done.

PR #180, opened 2021-04-26
C2LC-360: Remove fixed width for coordinate input text box and limit …

chosww:
…character input length to 2

### DIFF
--- a/src/ProgramBlockEditor.js
+++ b/src/ProgramBlockEditor.js
@@ -767,7 +767,7 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
                             className={characterPositionTextInputClassName}
                             aria-label={this.props.intl.formatMessage({id:'ProgramBlockEditor.editPosition.columnPosition'})}
                             aria-disabled={this.props.editingDisabled}
-                            maxLength='2'
+                            maxLength='1'
                             size='2'
                             type='text'
                             value={this.state.characterColumnLabel}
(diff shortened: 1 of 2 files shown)


chosww:
Hey @the-t-in-rtf, could you do some device testings for this PR please?

the-t-in-rtf:
This looks good on the simulated devices I've tried.  I tested the smallest iPad Pro (9.7") and iPad Air.  I saw the problem with [the live site](https://code.weavly.org/) on both devices at all levels of text zoom.  With your fix, I don't see the problem regardless of how far I zoom in and out.  So this looks good to me.

src/ProgramBlockEditor.js
```diff
@@ -767,6 +767,8 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
                             className={characterPositionTextInputClassName}
                             aria-label={this.props.intl.formatMessage({id:'ProgramBlockEditor.editPosition.columnPosition'})}
                             aria-disabled={this.props.editingDisabled}
+                            maxLength='2'
```
the-t-in-rtf, line 770: Is this just to make them match, or are you future-proofing?  We don't have any double character column positions.

the-t-in-rtf: I can see that having a `maxLength` at all is an improvement over what's deployed to production.  Can we try this with a `maxLength` of `1`?

chosww: Good idea, I changed maxLength of column text box to be 1, and left size to be 2 for its width.

the-t-in-rtf: With these changes I think this is ready for further review.

PR #181, opened 2021-04-26
C2LC-361: Implement a component to handle update of aria-live announcement of character position

chosww:


the-t-in-rtf:
See [https://issues.fluidproject.org/browse/C2LC-361](https://issues.fluidproject.org/browse/C2LC-361) for details.

src/ProgramBlockEditor.test.js
```diff
@@ -942,20 +928,3 @@ describe('Using change character position by column/row labels', () => {
         expect(characterYPositionCoordinateBox.props.onChange).toBe(undefined);
     });
 });
-
-describe('Character position gets updated on character-position div', () => {
```
the-t-in-rtf, line 946: If this is still worth testing (and it seems like it is), it should be moved to `App.test.js` instead of removing it.

the-t-in-rtf:
This works as expected with Chrome and VoiceOver.  I left one comment on a deleted test that I think should be preserved, otherwise this looks good to me.

We'll have to carefully coordinate the merge of this pull and #177, I expect there to be conflicts whichever way we do it.

src/ProgramBlockEditor.test.js
chosww: Yeah, I agree that the test is worth testing; however, we have only used App test file to see if the coding environment has rendered, so we haven't been writing test cases for other logics in App. I see that test cases for AudioImpl has added to it, and was wondering if we should write more test cases for App? What do you think @sbates-idrc?

sbates-idrc:
I'm thinking that rather than placing this functionality in `App`, we could put it into its own component. This will enable us to unit test and to reduce the amount of new code that we add to the `App` component. It won't have any React rendered content: we should be able to provide a `render()` method that returns `null`.

chosww:
Thanks @sbates-idrc for the review, I have implement a component that is responsible for updating the aria-live region and tested using VoiceOver.

PR #183, opened 2021-04-29
C2LC-283: Resize icon when command is added

chosww:


the-t-in-rtf:
I see the effect on Firefox and Safari, but on Chrome it either doesn't happen or happens too quickly.

the-t-in-rtf:
It actually seems OK on Opera but too fast on Opera.

the-t-in-rtf:
This still doesn't seem very smooth to me, it kind of lurches from keyframe to keyframe.  See animated gifs below.

First, Firefox:

Add Command Animation- FF

and Safari:

Add Command Animation- Safari

and Chrome:

Add Command Animation- Chrome

I checked out your work and tried removing the intermediate keyframes, which results in a much smoother transition:

Add Command Animation- Two Keyframes

If you have another reason to use the intermediate keyframes, you may need more of them or to adjust the maths for each step.

src/_variables.scss
```diff
@@ -42,3 +42,17 @@ $c2lc-action-panel-height: 3rem;
 $c2lc-scene-border-width: 0.1rem;
 $c2lc-scene-border-color: #CCE6E2;
 $c2lc-scene-border-style: solid;
+
+/* mixins */
+
+@mixin updatedProgramAnimation($color, $commandName, $theme) {
```
the-t-in-rtf, line 48: The mixin is really an impressive addition, I think we'll be able to clean up a lot of duplicate code in the themes when we're ready.

src/_variables.scss
```diff
@@ -42,3 +42,17 @@ $c2lc-action-panel-height: 3rem;
...
+    @keyframes updatedAnimation-#{$commandName}-#{$theme} {
+        0% {box-shadow: 0 0 0 1rem $color;}
+        25% {box-shadow: 0 0 0 0.75rem $color;}
```
the-t-in-rtf, line 51: As mentioned in my comment with gifs, the transition seemed much smoother without the 25/50/75 percent keyframes.  Was there another reason you needed them?

sbates-idrc:
I've committed some changes to the management of the animation class:

https://github.com/sbates-idrc/c2lc-coding-environment/commit/46a351dafc964a7893b3db329371179f08cf2b4f

Changes:

- Register an event handler for 'onAnimationEnd' and remove the "ProgramBlockEditor__program-block--updated" class when the animation has completed.
- Use a new ProgramBlockEditor instance variable 'updatedCommandBlockIndex' to manage which block should have the "ProgramBlockEditor__program-block--updated" class added to it, rather than reusing 'focusCommandBlockIndex'.

Please review my changes and merge into your C2LC-283 branch.

PR #166, opened 2021-04-12
C2LC-349: Fixed tuning of movement sounds to reflect scene dimensions.

the-t-in-rtf:
See [C2LC-349](https://issues.fluidproject.org/browse/C2LC-349) for details.

PR #185, opened 2021-04-30
C2LC-372: parameterised scene dimension minX and minY values.

the-t-in-rtf:
See [C2LC-372](https://issues.fluidproject.org/browse/C2LC-372) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -103,7 +103,7 @@ export class App extends React.Component<AppProps, AppState> {
             bluetoothApiIsAvailable: FeatureDetection.bluetoothApiIsAvailable()
         };
 
-        this.sceneDimensions = new SceneDimensions(26, 16);
+        this.sceneDimensions = new SceneDimensions(1, 26, 1, 16);
 
         // Begin facing East
         this.startingCharacterState = new CharacterState(1, 1, 2, [], this.sceneDimensions);
--- a/src/AudioManagerImpl.test.js
+++ b/src/AudioManagerImpl.test.js
@@ -48,7 +48,7 @@ test("Returns a sensible note range for every supported character position.", ()
     // noteTable [row][col] = singlePitchString;
     const noteTable = [];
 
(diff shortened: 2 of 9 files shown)


chosww:
Thanks @the-t-in-rtf, the changes you made looks good, and it makes sense to make adjustment of 0.5 on where it is needed. There's one thing you forgot to update, and that is how the scene is drawn in Scene component; the scene is using min/max of boundaries to draw the scene, and translated x and y pos should be used to draw the scene. Also, the path is drawn incorrectly as well, so after you update those, I will take another look.

the-t-in-rtf:
I updated the path drawing logic to also offset itself (and updated the associated tests).

the-t-in-rtf:
Daniel also mentioned that the last row/column of the scene grid weren't being drawn, I'm fixing that now.

the-t-in-rtf:
I fixed the outstanding issues I was aware of and updated with `develop-o.8`. There were (expectable) merge conflicts in updated tests that used the old syntax for creating scene dimensions.  There was also a test failure related to the previous change that I mistakenly thought was caused by the merge, but which is fixed in the last commit.

the-t-in-rtf:
The last error was a test coverage threshold error and not a test failure.  I'll look into that in a bit.

chosww:
Thanks @the-t-in-rtf, the changes look good, and draws the grid and path correctly. After the coverage issue is fixed, this PR is good for @sbates-idrc to review.

the-t-in-rtf:
As discussed in [C2LC-349](https://github.com/codelearncreate/c2lc-coding-environment/pull/166#issuecomment-833455971):

> The CI build failures were because of "branch" coverage changes, namely not reaching the conditional in the Scene that is configured for zero width and zero height scene dimensions. I fixed it by removing that conditional, which is unreachable because we cannot create a zero width or zero height scene using the update scene dimensions class.

PR #190, opened 2021-05-12
C2LC-333: Initial work on global keyboard handler and basic shortcuts.

the-t-in-rtf:
See [C2LC-333](https://issues.fluidproject.org/browse/C2LC-333) for details.


chosww:
Thanks @the-t-in-rtf for starting this work, I tried the initial implementation on Windows and MacOS in Chrome browser, and here are what didn't work:

- Robot Character position announcement is not working on MacOS (alt + i)
- Audio toggle can only turn the audio toggle off, and can't turn it back on
- Stop program seems not working on both Windows and MacOS
- ? key is not doing anything on both Windows and MacOS

This work definitely needs a lot of reviews and manual testings in different device combinations.

the-t-in-rtf:
The question mark key is not implemented yet, I suspect most of the rest broke when I refactored to avoid tracking the "alt" key manually in favour of using the `altKey` property of the keyboard event.

the-t-in-rtf:
> Audio toggle can only turn the audio toggle off, and can't turn it back on

I was trying to reuse the function we use with the onscreen toggle, but as the change is based on inverting the current state, we needed to use the function form, i.e. `setState((state)=>{ ... })` instead.  This fixes keyboard toggling of audio.

> Stop program seems not working on both Windows and MacOS

I had a typo in the `runningState` checks for that key binding, the second check should have used `!==`.  With that fixed, it works as expected.

> Robot Character position announcement is not working on MacOS (alt + i)

This works for me, but maybe there are some issues with different keyboard layouts.  If you add a breakpoint or debugger statement to the default case for both switch statements, you should be able to see what "your" alt+i is being detected as.  I also noticed that if I have the debugging console open, it often steals keyboard focus, you have to explicitly click somewhere in the app before any key will work.

chosww:
Thanks @the-t-in-rtf for the updates, the modal and styles you have used looks great! The issues I described in earlier comments got resolved, but it seems like short keys with alt or shift combination doesn't work with NVDA. Any key presses combined with alt or shift gets either have no effect or NVDA has built in shortkeys for them (alt + i for feedback form, alt + b moves to the bookmark tab).

the-t-in-rtf:
> * ? key is not doing anything on both Windows and MacOS

I just implemented the modal.  Even though it only displays the keys at the moment, I called it "keyboard input" as we may evolve it to allow configuration.

the-t-in-rtf:
Quick note, this now also includes the hotkeys to add the selected command to the beginning and end of the program.

src/KeyboardInputModal.js
```diff
@@ -49,15 +67,17 @@ class KeyboardInputModal extends React.Component<KeyboardInputModalProps, {}> {
...
+                aria-modal={true}
+                role="dialog"
                 dialogClassName='KeyboardInputModal'
             >
                 <Modal.Body className='KeyboardInputModal__content'>
                     <h2 className="KeyboardInputModal__content__title">
                         <FormattedMessage id='KeyboardInputModal.Title'/>
                     </h2>
-                    <div>
+                    <ul className="KeyboardInputModal__content__list">
```
chosww, line 78: Seems like default ul style has padding left caused by "padding-inline-start: 40px". Because of this style, the list items seems not horizontally centred. Let's have a css rule for className KeyboardInputModal__content__list to remove the padding.

in KeyboardInputModal style sheet, 
```
.KeyboardInputModal__content__list {
padding-inline-start: 0;
}
```

chosww:
Thanks @the-t-in-rtf, I tested it with NVDA, and I see a few issues. 

- The modal's list and its items are not perceivable by NVDA; when I press L, NVDA announces, "No List Found" and when I press I, NVDA announces, "No List Item Found". 
- The modal contents are not perceivable by NVDA; using navigation key only announces black, and space when the modal is open. DeleteAll Modal is working fine, so maybe a good idea to check what's different.
- Alt shortcuts are working fine, but alt shortcuts may cause an issue on Chrome browser, as pressing alt key alone will put focus to the menu tab on the tab bar and need to click back on the application to interact with it.
- Shift key is still not working, but apparently alt + shift + ? opens up the modal, and can use speed controller shortcuts. I am not sure, but NVDA may need to press alt key with shift key.

src/KeyboardInputSchemes.js
```diff
@@ -0,0 +1,128 @@
...
+        return "";
+    }
+};
+
+export function keyboardEventMatchesKeyDef (e: KeyboardEvent, keyDef: KeyDef) {
+    if (e.code === keyDef.code || e.key === keyDef.key) {
+        if (!!(keyDef.altKey) !== !!(e.altKey)) {
+            return false;
+        }
+        if (!!(keyDef.shiftKey) !== !!(e.shiftKey)) {
```
chosww, line 118: Any key bindings with shift key gets filtered by this check, because keyDef.shiftKey is undefined for "<", ">" and "?", which result in those key bindings to not work. I was thinking maybe we could make keyDefs with code + shiftKey combination?

the-t-in-rtf: I agree, and this explains recent issues getting the bindings to work as well. When I added tests, I hardened the checks to fix false matches (triggering actions when the right keys weren't hit). That seems to have resulted in the behaviour you picked up on.

the-t-in-rtf: I ended up relaxing this check, my reasoning is that since caps lock is often used as a screenreader activation key, we may not want to be as strict about ctrl-a vs ctrl-A.   We should discuss in the meeting, it may make more sense not to include / rely on the shift state at all in key bindings.

src/KeyboardInputModal.js
the-t-in-rtf: I did this a while back, but ended up removing it later in favour of an explicit `padding-left` on the content list once I got rid of the padding inherited from the bootstrap modal styles.

src/App.js
```diff
@@ -483,6 +499,7 @@ export class App extends React.Component<AppProps, AppState> {
                 break;
             case 'stopRequested': // Fall through
             case 'stopped':
+                this.handleClickStop();
```
chosww, line 502: Case 'stopped' is when program is not running, so calling handleClickStop doesn't make sense over here, as what handleClickStop does is setting runningState to 'stopRequested'. Is this call intended?

src/KeyboardInputModal.scss
```diff
@@ -0,0 +1,164 @@
...
+}
+
+.KeyboardInputModal__binding__icon {
+    border: 2px solid #B9BEC6;
+    background-color: white;
+    border-radius: 3px;
+    box-shadow: 0 3px 1px 0 #B9BEC6;
+    font-weight: bold;
+    height: 2.5rem;
+    padding-top: 0.25rem;
```
chosww, line 117: padding-top pushes the icons to look like vertically centred, but it is not exactly centred. There could be other ways, but I found using flex to center things around convenient. Here is a suggestion, and note that I removed padding-top and text-align.
 
```
.KeyboardInputModal__binding__icon {
display: flex;
justify-content: center;
align-items: center;
border: 2px solid #B9BEC6;    
background-color: white;    
border-radius: 3px;    
box-shadow: 0 3px 1px 0 #B9BEC6;    
font-weight: bold;    
height: 2.5rem;
width: 2.5rem;
}
```

src/KeyboardInputModal.scss
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+.KeyboardInputModal__content__list--disabled {
+    display: none;
+}
+
+.KeyboardInputModal__binding {
+    display: grid;
+    grid-template-columns: 6rem auto;
```
chosww, line 98: 6rem is not enough width to render 3 key combinations, which result them to look squished. Each icon is 2.5rem width, so we would need to change this value to 5rem on 2 key combinations layout and 7.5rem on 3 key combinations. Setting this to min-contnet would be much easier, but we have ? icon, which doesn't have same number of icons in a row as other commands.

src/KeyboardInputModal.js
```diff
@@ -0,0 +1,216 @@
...
+                    </div>
+
+                    {this.renderKeyboardSchemeMenu()}
+
+                    <ul className={"KeyboardInputModal__content__list" + (this.props.keyBindingsEnabled ? "": " KeyboardInputModal__content__list--disabled")}>
+                        {this.renderKeyBindings()}
+                    </ul>
+
+                    <div className="KeyboardInputModal__content__footer">
+                        <Button className="KeyboardInputModal__content__cancelButton" onClick={this.props.onHide}>
```
chosww, line 194: Both "Cancel" and "Done" buttons are functioning the same on click, could you confirm if that's intended with Sepideh please? If they are the same, I am not sure why we are using two buttons for that.

the-t-in-rtf: I'll ask her, I had the same thought but wanted to start with what she'd designed.

src/App.js
the-t-in-rtf: It doesn't seem appropriate, and the behaviour is as expected without it, so I removed it.

src/KeyboardInputModal.scss
the-t-in-rtf: Thanks, I had to fudge it and knew there had to be a better way.

src/KeyboardInputModal.js
the-t-in-rtf: I talked to Sepi, she wants the "Done" button to actually make the state changes, and "Cancel" to abandon them.  This will require a light refactor but isn't difficult.

src/KeyboardInputModal.scss
the-t-in-rtf: I ended up making the column 7.5rem for now, and it looks better.  There is a small concern if we start supporting 4-key combos (`shift+ctrl+alt+X`), but we can discuss that when we have to.

src/KeyboardInputModal.js
the-t-in-rtf: I rearranged things so that:

1. Changes are displayed onscreen as they're made.
2. Changes are passed on to the app when "Done" is hit.
3. Changes are reset when "Cancel" is hit or the dialogue is closed without hitting "Done".

src/App.js
```diff
@@ -606,69 +608,131 @@ export class App extends React.Component<AppProps, AppState> {
...
+                                window.speechSynthesis.speak(utterance);
+                            }
+                            break;
+                        case("playPauseProgram"):
+                            if (this.state.programSequence.getProgramLength() > 0) {
+                                this.handleClickPlay();
+                            }
+                            break;
+                        case("refreshScene"):
+                            if (this.state.runningState !== 'running') {
```
chosww, line 663: We want more strict condition check here, as we want to make sure any program execution has stopped. Also, to match when we are disabling the RefreshButton, let's change the condition to match RefreshButton's disabled condition check.

```javascript
if ( !(this.state.runningState === 'stopped' || this.state.runningState === 'paused') ) 
```

chosww:
Thanks @the-t-in-rtf, the changes look good and I didn't find any issues with the code as well as from some manual testings.

PR #191, opened 2021-05-18
C2LC-393: Removed world icon dark background from 'mixed' theme.

the-t-in-rtf:
See [C2LC-393](https://issues.fluidproject.org/browse/C2LC-393) for details.


chosww:
This change looks good and removes the background colour for the world selectors as the design on Figma.

PR #196, opened 2021-06-07
C2LC-376: Update colours of forward and backward commands

chosww:
See [C2LC-376](https://issues.fluidproject.org/browse/C2LC-376) for details.


the-t-in-rtf:
The code changes are trivial, nothing stood out to me.

I tested this briefly to confirm that the change is there and looks good in Mixed (default) and Dark, and when combined with the styles for the Light theme.  I also checked the other themes to confirm that no updated styles were inherited inappropriately.

This seems ready for further review.

PR #199, opened 2021-06-09
C2LC-389: Make Weavly heading to direct to weavly.org

chosww:
See [C2LC-389](https://issues.fluidproject.org/browse/C2LC-389) for details.


src/App.js
```diff
@@ -727,7 +727,9 @@ export class App extends React.Component<AppProps, AppState> {
                     <header className='App__header'>
                         <div className='App__header-row'>
                             <h1 className='App__app-heading'>
-                                <FormattedMessage id='App.appHeading'/>
+                                <a href='https://weavly.org'>
```
the-t-in-rtf, line 730: This should probably include `target="_blank"`.  I just tested with Chrome to confirm, if I change a non-persisted setting (for example, toggling the audio), it's not retained when I go back to the Coding Environment using the back button.

chosww: Thanks @the-t-in-rtf, that's a good idea.

the-t-in-rtf:
With the changes to the link attribute, this seems ready for further review.

PR #200, opened 2021-06-09
C2LC-391: Fix AddNode button's svg gets shifted when focused

chosww:
See [C2LC-391](https://issues.fluidproject.org/browse/C2LC-391) for details.


src/Themes.scss
```diff
@@ -242,7 +242,7 @@ body.light-theme {
 
     button.btn.AddNode__expanded-button {
         background-color: white;
-        border: 3px solid #505862;
+        border: 2px solid #505862;
```
the-t-in-rtf, line 245: Why not use the `addNodeExpandedButtonBorderWidth` variable here?

the-t-in-rtf:
This addresses the issue across all themes.  I had one minor comment on a hard-coded width that seems to match an existing variable.

src/Themes.scss
chosww: Currently, addNodeExpandedButtonBorderWidth is defined in AddNode.scss and from the theming work, any border for the expanded add node button are hardcoded in Theme.scss.

chosww: I am thinking, we should merge this PR with this change, and have a clean up task to move any commonly used values in different themes to variables.scss, what do you think?

the-t-in-rtf: That sounds reasonable.  We could also consider making some more animations into mixins.

chosww: This work is ready for a review @sbates-idrc, please let me know what do you think about the comments Tony and I left.

PR #201, opened 2021-06-14
C2LC-381: Move the character positioning controls to the world selector panel

chosww:
See [C2LC-381](https://issues.fluidproject.org/browse/C2LC-381) for details.


the-t-in-rtf:
This looks fine except for the light theme.  The grey behind the controls is darker than the grey behind the character column of the program editor, and it doesn't seem like the contrast is good enough.  I'd suggest reviewing it with Sepi.

the-t-in-rtf:
By contrast I mean there isn't enough difference between the hover controls and the normal style in the grayscale theme.

chosww:
New hover colour for the reposition buttons in the grayscale theme: #3A3F46

sbates-idrc:
The `App.handleChangeCharacterPosition` method should use the function form of `setState` as the state change is dependent on the current state.

src/CharacterPositionController.js
```diff
@@ -0,0 +1,234 @@
...
+                prevPropsCharacterState: currentCharacterState,
+                characterColumnLabel: currentCharacterState.getColumnLabel(),
+                characterRowLabel: currentCharacterState.getRowLabel()
+            };
+        } else {
+            return null;
+        }
+    }
+
+    handleClickCharacterPosition = (e) => {
```
sbates-idrc, line 58: We can name the methods of this component more consistently and more clearly to better communicate their responsibility and relationship:

* `handleClickCharacterPosition` -> `handleClickCharacterPositionButton`
* `handleKeyDownCharacterPosition` -> `handleKeyDownCharacterPositionButton`
* `handleChangeCharacterPosition` -> I think we can just remove this method and call `this.props.onChangeCharacterPosition` directly in the two methods above
* `handleChangeCharacterPositionLabel` -> fine as is
* `handleBlurCharacterPositionBox` -> `handleBlurCharacterPositionLabel`
* `handleUpdateCharacterPosition` -> `handleKeyDownCharacterPositionLabel`

src/CharacterPositionController.js
```diff
@@ -0,0 +1,230 @@
...
+                prevPropsCharacterState: currentCharacterState,
+                characterColumnLabel: currentCharacterState.getColumnLabel(),
+                characterRowLabel: currentCharacterState.getRowLabel()
+            };
+        } else {
+            return null;
+        }
+    }
+
+    handleClickCharacterPositionButton = (e) => {
```
sbates-idrc, line 58: Please add a type for the 'e' parameter.

src/CharacterPositionController.test.js
```diff
@@ -0,0 +1,220 @@
...
+        const sampleXPosition = 'X';
+        const secondSampleXPosition = 'A';
+        const eventObject = (value) => (
+            {
+                key: 'Enter',
+                preventDefault: () =>  {},
+                currentTarget:
+                    { name: 'xPosition', value }
+            }
+        );
```
sbates-idrc, line 151: I think this function is a good basis for making some general test utility functions. I'd suggest doing the following (and also removing the equivalent one in the `'Changing y position'` test):

- Split into 3 functions:
  - `makeChangeEvent()`
  - `makeBlurEvent()`
  - `makeKeyDownEvent()`
- Move them to a new `TestUtils.js` file so that they can be used in other tests

chosww: Thanks for the comment @sbates-idrc, I see that making TestUtils would be helpful, but this case is specific to the CharacterPositionCoordinateBox, so it has specific values hard coded like key: 'Enter', and currentTarget object with name 'xPosition'. Are you thinking about specifying the hardcoded values as parameters to the functions?

PR #202, opened 2021-06-14
C2LC-397: Change the label for the add node when adding to an empty program

chosww:
See [C2LC-397](https://issues.fluidproject.org/browse/C2LC-397) for details.


the-t-in-rtf:
The code changes are minimal and this works as expected, the announcement when the keyboard focus moves to the first node is correct with VoiceOver both when the program is empty and when there are existing steps.

the-t-in-rtf:
@sbates-idrc, this is ready for further review.

PR #205, opened 2021-06-23
C2LC-398: Fixed hover styles for share button in light theme.

the-t-in-rtf:
See [C2LC-398](https://issues.fluidproject.org/browse/C2LC-398) for details.

### DIFF
--- a/src/ShareButton.scss
+++ b/src/ShareButton.scss
@@ -1,6 +1,6 @@
 @import '_variables';
 
-button.btn.ShareButton {
+.ShareButton {
     background-color: #4C9990;
     border: 0;
     color: #EBFAF8;
--- a/src/ShareButton.scss
+++ b/src/ShareButton.scss
@@ -1,6 +1,6 @@
 @import '_variables';
 
-.ShareButton {
+button.btn.ShareButton {
     background-color: #4C9990;
     border: 0;
     color: #EBFAF8;
--- a/src/Themes.scss
+++ b/src/Themes.scss
@@ -296,8 +296,10 @@ body.light-theme {
     }
 
     button.btn.ShareButton {
-        background-color: #F1F2F4;
         color: black;
+        &:not(:hover) {
+            background-color: #F1F2F4;           
+        }
     }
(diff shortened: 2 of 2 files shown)


src/Themes.scss
```diff
@@ -296,8 +296,10 @@ body.light-theme {
     }
 
     button.btn.ShareButton {
-        background-color: #F1F2F4;
         color: black;
+        &:not(:hover) {
```
chosww, line 300: Setting more specific rule for the base background-color seems not a good approach. We could have less specific rule for the ShareButton, so hover style can be applied from ShareButton component -- removing button.btn in front of the .ShareButton and setting background-color how it was before would fix this issue.

chosww:
```
    .ShareButton {
        color: black;
        background-color: #F1F2F4;
    }
```

chosww:
I also noticed that share button's hover colour seems not right in the grayscale theme. I assume that the correct colour should be the darker hover colour on other components, but would be if we could ask Sepideh to confirm that.

src/Themes.scss
the-t-in-rtf: I fixed this instance of it, but we still have lots of examples from the earlier theme work.  We can discuss a wider CSS cleanup in another ticket, or do it incrementally as we improve other elements.

the-t-in-rtf:
I used the hover styles for the share button in the grayscale theme to match the theme selector, which looks reasonable.  Feel free to share with Sepi to confirm.

src/ShareButton.scss
```diff
@@ -1,6 +1,6 @@
 @import '_variables';
 
-button.btn.ShareButton {
```
chosww, line 3: We still need button.btn part to apply the default interaction styles to the light theme. With this removed,the  light theme has no hover style.

the-t-in-rtf: Good catch, and fixed.

chosww:
Thanks @the-t-in-rtf, the changes look good.

PR #206, opened 2021-06-23
C2LC-401: Add auditory preview for focus

chosww:
See [C2LC-401](https://issues.fluidproject.org/browse/C2LC-401) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -590,12 +590,15 @@ export class App extends React.Component<AppProps, AppState> {
         this.focusTrapManager.handleKeyDown(e);
     };
 
-    handleRootFocus = (e) => {
+    handleRootFocus = (e: SyntheticKeyboardEvent<HTMLElement>) => {
+        // $FlowFixMe classList is missing in event target
         if (e.target.classList.contains('focus-announcement')) {
+            // $FlowFixMe getAttribute is missing in event target
             const ariaLabel = e.target.getAttribute('aria-label');
             if (ariaLabel) {
                 this.audioManager.playStringMessage(ariaLabel);
             } else {
+                // $FlowFixMe innerText is missing in event target
                 const innerText = e.target.innerText;
                 if (innerText) {
                     this.audioManager.playStringMessage(innerText);
--- a/src/App.js
+++ b/src/App.js
@@ -590,6 +590,20 @@ export class App extends React.Component<AppProps, AppState> {
         this.focusTrapManager.handleKeyDown(e);
     };
 
+    handleRootFocus = (e) => {
+        if (e.target.classList.contains('focus-announcement')) {
+            const ariaLabel = e.target.getAttribute('aria-label');
+            if (ariaLabel) {
+                this.audioManager.playStringMessage(ariaLabel);
+            } else {
+                const innerText = e.target.innerText;
+                if (innerText) {
+                    this.audioManager.playStringMessage(innerText);
+                }
+            }
+        }
+    }
+
     handleToggleAudioFeedback = (announcementsEnabled: boolean) => {
         this.setState({
             announcementsEnabled: announcementsEnabled
@@ -723,7 +737,8 @@ export class App extends React.Component<AppProps, AppState> {
                     className='App__container'
                     role='main'
                     onClick={this.handleRootClick}
-                    onKeyDown={this.handleRootKeyDown}>
+                    onKeyDown={this.handleRootKeyDown}
+                    onFocus={this.handleRootFocus}>
                     <header className='App__header'>
                         <div className='App__header-row'>
                             <h1 className='App__app-heading'>
--- a/src/AudioManagerImpl.js
+++ b/src/AudioManagerImpl.js
@@ -233,6 +233,16 @@ export default class AudioManagerImpl implements AudioManager {
         }
     }
 
+    playStringMessage(message: string) {
+        if (this.announcementsEnabled) {
+            if (window.speechSynthesis.speaking || window.speechSynthesis.pending) {
+                window.speechSynthesis.cancel();
+            }
+            const utterance = new SpeechSynthesisUtterance(message);
+            window.speechSynthesis.speak(utterance);
+        }
+    }
+
     // TODO: Add a better type for pitch.
     // TODO: Make this private, as it doesn't respect the audioEnabled setting.
     playPitchedSample(sampler: Sampler, pitch: string, releaseTime: number) {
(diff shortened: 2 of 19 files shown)


the-t-in-rtf:
The preview for the add node button doesn't seem to match the rest.  All of the other announcements are telling you what you're focused on, and not what will happen if you click them.  The add node button is telling you what will happen if you click the control, in this case a reason why it can't happen.

To make them match, first I'd suggest improving the other preview announcements.  As it stands, it's a little like a "you are here" sign without the "you are" part. Something as simple as "Open Theme Menu" or "Disable announcements", would be more of a "preview".  

Not sure how to resolve the add node issue, perhaps "Add Action is currently not available because no command is selected."?

src/AudioManagerImpl.js
```diff
@@ -233,6 +233,16 @@ export default class AudioManagerImpl implements AudioManager {
         }
     }
 
+    playStringMessage(message: string) {
```
the-t-in-rtf, line 236: This is very close in function to `playAnnouncement`, with quite a bit of duplicated code.  I would suggest refactoring `playAnnouncement` to just look up the internationalised content and have it use `playStringMessage` to check `announcementsEnabled` and play the actual announcement if needed.

the-t-in-rtf: It might also be good to make `playAnnouncement` something like `playMessageId` to more closely match its new function.

chosww: Thanks for the comment @the-t-in-rtf, that makes sense to refactor playAnnouncement and use playStringMessage to play the announcement. About the other suggestions, we would need two separate functions for the announcements, one for preview and one for feedback, so I am thinking changing the name to playFeedbackAnnouncement and have the preview one named playPreviewAnnouncement would make more sense.

the-t-in-rtf:
I'm happy for this to move on to a full review if you are, @chosww.

chosww:
Thanks for your reviews @the-t-in-rtf, I think we don't have much time left for 0.9 release, so your comment about the preview of AddNode should be recorded somewhere else and get some designers feedback later. I also think that there was no feedback around that during the Fable user testing sessions, so it might be ok as it is, but worth discuss about it with Sepideh and maybe Lisa.

the-t-in-rtf:
Given that you're just reusing the existing ARIA labels, I agree that the wording improvements can wait.

src/ProgramBlockEditor.js
```diff
@@ -133,7 +133,13 @@ class ProgramBlockEditor extends React.Component<ProgramBlockEditorProps, Progra
...
         // Play the announcement
         const commandString = this.props.intl.formatMessage({ id: "Announcement." + this.props.programSequence.getProgramStepAt(index)});
-        this.props.audioManager.playAnnouncement('delete', this.props.intl, { command: commandString});
+        this.props.audioManager.playFeedbackAnnouncement(
+            this.props.intl.formatMessage(
+                {id: 'Announcement.delete'},
+                {command: commandString}
+            )
+        );
+        this.props.audioManager.setFeedbackIsPlaying(true);
```
the-t-in-rtf, line 142: It seems like this should be set in `playFeedbackAnnouncement`?

src/App.js
```diff
@@ -854,14 +854,18 @@ export class App extends React.Component<AppProps, AppState> {
...
+                this.audioManager.playPreviewAnnouncement(ariaLabel);
+            } else {
+                // $FlowFixMe innerText is missing in event target
+                const innerText = e.target.innerText;
+                if (innerText) {
+                    this.audioManager.playPreviewAnnouncement(innerText);
+                }
             }
+        } else {
+            this.audioManager.setFeedbackIsPlaying(false);
```
the-t-in-rtf, line 868: I still think the audio manager could handle both parts of this, i.e. you should always be able to ask it to announce something and let it figure out what to do if it's already playing.

src/AudioManagerImpl.js
```diff
@@ -232,6 +232,7 @@ export default class AudioManagerImpl implements AudioManager {
                 window.speechSynthesis.cancel();
             }
             this.playStringMessage(message);
```
the-t-in-rtf, line 234: Why wouldn't you set the flag in `playStringMessage`?  In other words, are there any circumstances in which you'd play feedback without setting the flag?

the-t-in-rtf: In reading the code, it seems like the intent of the flag is to avoid event attempting to play a preview if feedback is already being announced?  Is this correct?

src/App.js
```diff
@@ -851,6 +851,24 @@ export class App extends React.Component<AppProps, AppState> {
...
+                this.audioManager.playPreviewAnnouncement(ariaLabel);
+            } else {
+                // $FlowFixMe innerText is missing in event target
+                const innerText = e.target.innerText;
+                if (innerText) {
+                    this.audioManager.playPreviewAnnouncement(innerText);
+                }
+            }
+        } else {
+            this.audioManager.setFeedbackIsPlaying(false);
```
the-t-in-rtf, line 868: This seems incorrect.  If a feedback announcement is playing, the first pass through this will clear the flag, so depending on the length of the feedback announcement, you might still attempt to play a preview announcement anyway.

I'd suggest setting the flag to false using the [`end` event](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisUtterance/end_event) for an utterance instead.  One way to handle this would be to support an optional callback (to clear the event) as a second argument to `playStringMessage`.

src/AudioManagerImpl.js
chosww: No, the intention is to prevent preview announcement played by focus when interacting with components, if there is a feedback announcement. Without this flag, feedback announcement about program step added and deleted is not played because of the preview announcement. While feedback announcement is playing, another feedback announcement can stop that and played.

src/App.js
chosww: That is exactly what I want; clear the flag during the first pass through. I think feedback and preview announcement have equal importance, so any of the two announcements can cancel each other.

src/AudioManagerImpl.js
the-t-in-rtf: Then it definitely still needs work, as you could trivially clear the flag by changing focus rapidly. See my other note for a suggested fix.

src/App.js
the-t-in-rtf: OK, thank you for explaining further, maybe a comment above the check explaining the rationale is enough?  i.e. without the flag/check we do X too often, et cetera.

the-t-in-rtf:
I did some QA to see if I could get it to skip announcements or make double announcements, and it works as well as everything else we have.  I do think some light renaming or comments in the code would be tidier, but this does seem to work as expected.

the-t-in-rtf:
I tested this with Chrome and Safari, and it works as expected.  I used the combination of tab and spaces in the action panel to confirm that:

1. The last preview (hover) announcement is played after the feedback announcement.
2. Preview announcements interrupt each other as expected.
3. Feedback announcements interrupt each other as expected.

This seems ready for further review.

sbates-idrc:
Weavly has changed since the initial auditory preview work. If we do auditory preview in the future, we will create new designs and issues at that point. Closing the existing ones.

PR #208, opened 2021-07-05
C2LC-403: Add a UI for turning focus and hover auditory preview on and off

chosww:
See [C2LC-403](https://fluidproject.atlassian.net/browse/C2LC-403) for details.


the-t-in-rtf:
After working with this for a while, I believe I understand the intent, but I think it would be helpful to confirm that by documenting the goal of this change on the JIRA ticket. My understanding is that "preview" announcements are indicating focus, and "feedback" announcements are indicating (some) state changes.  Is this accurate?

If the intent with the preview announcements is to reinforce what's focused, I think the wording needs to be slightly clearer.  Perhaps just adding the word "highlighted" or "has focus" to the preview announcements would be enough?  This would also make it clearer what's happening when you reenter the window.  Try this with preview announcements on:

1. Highlight any element, a preview announcement is played.
2. Open or navigate to another window.
3. Come back to the window, the preview announcement is played again.

With a wording change, this would make sense, and would  more closely match what VoiceOver does in the same scenario.  It would also make it much clearer that you're not necessarily changing the state of the pen toggle to "pen down" when you click it. 

I also think we need to discuss and write down how they should interoperate when both are enabled.  Right now the behaviour seems confusing.  When adding a new command, replacing a command, or deleting a command, only the preview announcement is made.  When selecting a command, only the feedback announcement is made.  

It's particularly confusing when using the action panel to delete a step, which announces the focus on the next step, and never announces that a step was deleted.  We should be consistent in the cases where there's overlap.  One approach would be to announce state changes (feedback), and then announce the focus (preview).

I also wanted to ask about the decision to move away from a modal.  This leaves us without a toggle for movement sounds.  As a result, we are always making noise over the top of both "our" announcements and anything a screen reader is trying to convey.  I don't think this is ideal.  I also think the modal would make it much easier to convey the intent of the individual toggles.  The icons don't really indicate what each does (I had to inspect the document's markup to figure it out).  With a modal, you could use short descriptions to make things clear.

I also think we should briefly discuss what state changes (feedback) we're announcing.  Currently we announce a change in selected command, when a command is added, replaced, and deleted.  Why do we announce the things we do and not the others?  If our goal is to provide a screen reader like experience without requiring a screen reader, we might want to create a few additional issues to close some gaps, like adding preview announcements for position changes, world changes, theme changes, pen state, and expanded node state.

src/AudioManagerImpl.js
```diff
@@ -221,18 +222,25 @@ export default class AudioManagerImpl implements AudioManager {
...
-            if (window.speechSynthesis.speaking || window.speechSynthesis.pending) {
-                window.speechSynthesis.cancel();
-            }
-            const messageId = "Announcement." + messageIdSuffix;
-            const toAnnounce = intl.formatMessage({ id: messageId}, messagePayload);
-            const utterance = new SpeechSynthesisUtterance(toAnnounce);
-            window.speechSynthesis.speak(utterance);
+    playFeedbackAnnouncement(message: string) {
+        if (this.audioFeedbackEnabled) {
+            this.cancelSpeech();
```
the-t-in-rtf, line 227: I would suggest canceling the speech in `playStringMessage` instead of doing it in each of the functions that call it.

src/AudioManagerImpl.js
```diff
@@ -221,18 +222,25 @@ export default class AudioManagerImpl implements AudioManager {
...
+            this.playStringMessage(message);
+        }
+    }
+
+    playPreviewAnnouncement(message: string) {
+        if (this.audioPreviewEnabled) {
+            this.cancelSpeech();
+            this.playStringMessage(message);
```
the-t-in-rtf, line 235: I understand that the aria labels are being reused here.  I would suggest using an i18n template for preview announcements, something like `"{arialLabel} has focus"` or `{ariaLabel} highlighted`.

src/AudioManagerImpl.js
```diff
@@ -221,18 +222,25 @@ export default class AudioManagerImpl implements AudioManager {
...
         }
     }
 
+    playStringMessage(message: string) {
+        const utterance = new SpeechSynthesisUtterance(message);
+        window.speechSynthesis.speak(utterance);
```
the-t-in-rtf, line 241: In the main conversation I commented on cases in which we'd want both feedback (state change) and preview (focus change) announcements.  Given that these are independently triggered, one option is to make some kind of queue for utterances and regularly check to see if we're still speaking or preparing to speak.  If we're no longer speaking and there are still unplayed utterances, pull the next one out of the queue.

src/App.js
```diff
@@ -380,7 +381,8 @@ export class App extends React.Component<AppProps, AppState> {
             selectedAction: null,
             isDraggingCommand: false,
             audioEnabled: true,
```
the-t-in-rtf, line 383: We should talk about what this variable has meant and should mean moving forward.  Some of the designs had an "all sounds" toggle as well as a "sound effects" toggle, and this most closely corresponds to the latter.  If we end up using a modal and having those two controls, I would suggest adding `audioEffectsEnabled` as well, and using `audioEnabled`  for "all sounds".

src/AudioManagerImpl.js
```diff
@@ -221,18 +222,25 @@ export default class AudioManagerImpl implements AudioManager {
...
-        if (this.announcementsEnabled) {
-            if (window.speechSynthesis.speaking || window.speechSynthesis.pending) {
-                window.speechSynthesis.cancel();
-            }
-            const messageId = "Announcement." + messageIdSuffix;
-            const toAnnounce = intl.formatMessage({ id: messageId}, messagePayload);
-            const utterance = new SpeechSynthesisUtterance(toAnnounce);
-            window.speechSynthesis.speak(utterance);
+    playFeedbackAnnouncement(message: string) {
+        if (this.audioFeedbackEnabled) {
```
the-t-in-rtf, line 226: This should also check `audioEnabled`.

src/AudioManagerImpl.js
```diff
@@ -221,18 +222,25 @@ export default class AudioManagerImpl implements AudioManager {
...
+            this.cancelSpeech();
+            this.playStringMessage(message);
+        }
+    }
+
+    playPreviewAnnouncement(message: string) {
+        if (this.audioPreviewEnabled) {
```
the-t-in-rtf, line 233: This should also check `audioEnabled`.

src/AudioManagerImpl.js
chosww: audioEnabled has a default state value of true in App.js, and we currently have no way to change the audioEnabled value. We should talk about what this state is for as you talked about in one of your comments.

src/App.js
chosww: I agree that we should talk about this. My understanding is that audioEnabled is for "all sounds"

src/AudioManagerImpl.js
chosww: That would cause a huge delay in the announcements when users are rapidly tabbing through the components. I think it makes sense to play one announcement at a time, regardless of the announcement type.

src/AudioManagerImpl.js
chosww: Good idea 👍

sbates-idrc:
Weavly has changed since the initial auditory preview work. If we do auditory preview in the future, we will create new designs and issues at that point. Closing the existing ones.

PR #210, opened 2021-07-08
C2LC-411: Add keyboard shortcut for delete program step

chosww:
See [C2LC-411](https://issues.fluidproject.org/browse/C2LC-411) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -661,7 +661,7 @@ export class App extends React.Component<AppProps, AppState> {
                         case("deleteCurrentStep"):
                             const currentElement = document.activeElement;
                             // $FlowFixMe: Not all elements have dataset property
-                            const index = parseInt(currentElement.dataset.stepnumber);
+                            const index = parseInt(currentElement.dataset.stepnumber, 10);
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -658,6 +658,20 @@ export class App extends React.Component<AppProps, AppState> {
                                 this.handleProgramSequenceChange(newProgramSequence);
                             }
                             break;
+                        case("deleteCurrentStep"):
+                            const currentElement = document.activeElement;
+                            // $FlowFixMe: Not all elements have dataset property
+                            const index = parseInt(currentElement.dataset.stepnumber);
```
sbates-idrc: Add-nodes have a `stepnumber` data attribute also. So this test isn't enough to safely determine if the active element is a program step. Let's add a new data attribute that specifies the element type called "`data-controltype`". For program steps, use:

    data-controltype='programStep'

And for add-nodes use:

    data-controltype=`addNode`

And then in this `deleteCurrentStep` handler, check to see if the `activeElement` has a data attribute `controltype` equal to `programStep`.

PR #211, opened 2021-07-08
C2LC-412: Add keyboard shortcuts to move a program step within a program

chosww:
See [C2LC-412](https://issues.fluidproject.org/browse/C2LC-412) for details.


the-t-in-rtf:
I tested this in both Chrome and Safari.  It works as expected.  I also tried it when unfocused and when focused on add node.  It moves the focused action when it should, and doesn't move anything when it shouldn't.  Seems ready for further review.

sbates-idrc:
Please update the target for this PR to `develop-1.3` and merge in branch `develop-1.3` to bring it up to date. Then please refactor the changes to use the `ProgramChangeController`.

docs/keyboard.md
```diff
@@ -66,6 +66,8 @@ follows.
 | Ctrl + Alt + x, f, x | focusCharacterColumnInput |
 | Ctrl + Alt + x, f, y | focusCharacterRowInput |
 | Ctrl + Alt + x, f, w | focusWorldSelector |
+| Ctrl + Alt + x, s, l | swapCurrentStepWithPreviousStep |
```
sbates-idrc, line 69: Maybe 'm' for "move" rather than 's'. As "move" is what we use in the UI announcements.

chosww:
Thanks for the heads up @the-t-in-rtf. @sbates-idrc I worked on refactoring the swapStep operation, and I am getting a warning that says "An update (setState, replaceState, or forceUpdate) was scheduled from inside an update function." and seems like that's caused by changing state of ProgramBlockEditor in ProgramChangeController's this.app.setState. I think it would be good to chat about ways to resolve this on Monday.

chosww:
@sbates-idrc I added some test cases for the new moveProgramStep method, but I couldn't figure out how to test all the branches, as the condition checks for them are using actionPanelStep state in App and I wasn't sure how to update state of App for the test.

the-t-in-rtf:
As a side effect of our permissive handling of lower case keys (without shift) and upper case keys (with shift), `ctrl+alt+{` now does the same as `ctrl+alt+[`.  If this is no longer desired, adding `shiftKey: false` to the key definition should fix this.

the-t-in-rtf:
Looks like we removed the code that distinguishes `shiftKey`-modified key presses, we'd have to put that back.

PR #215, opened 2021-07-12
C2LC-416: Escape out of partially entered keyboard shortcuts.

the-t-in-rtf:
See [C2LC-416](https://issues.fluidproject.org/browse/C2LC-416) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -624,6 +624,11 @@ export class App extends React.Component<AppProps, AppState> {
     // TODO: Convert to use keyboardEventMatchesKeyDef for each command in turn.
     handleDocumentKeyDown = (e: KeyboardEvent) => {
         if (this.state.keyBindingsEnabled) {
+            if (e.key === 'Escape') {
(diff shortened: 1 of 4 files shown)


chosww:
With this approach, escape gets added to sequenceInProgress, I think we can just use a check in handleDocumentKeyDown for the escape key and set sequenceInProgress to an empty array, and I think that is more reasonable solution to the description in the ticket. I suggest start from dev-0.9 branch and just add a check like this in handleDocumentKeyDown:

```javascript
if (e.key === 'Escape') {
  this.sequenceInProgress = [];
  return;
}
```

the-t-in-rtf:
It gets added initially,  but is cleared out once `findKeyboardEventSequenceMatches` returns false, otherwise it wouldn't work.  I agree that the approach you suggest is probably cleaner, though, and will update.

chosww:
Thanks @the-t-in-rtf, the changes look good, and test case you added looks good too.

PR #216, opened 2021-07-12
C2LC-413: Add a keyboard shortcut to add a program step

chosww:
See [C2LC-413](https://issues.fluidproject.org/browse/C2LC-413) for details.


the-t-in-rtf:
I think you need to resolve the conflicts with the refreshed keyboard docs pull that Simon just merged, shouldn't be hard.

In testing this, I see that you made sure that you could repeatedly hit the add shortcut and have it keep going.  From the discussion, it sound like we agreed that it should be possible to add when either an existing step or an add node is focused, and that seems to work fine.

src/KeyboardInputSchemes.test.js
```diff
@@ -53,7 +53,7 @@ it('Should be able to handle a complete valid sequence', () => {
 
 it('Should be able to handle a complete invalid sequence', () => {
     const completeInvalidSequence = [
-        new KeyboardEvent('keydown', { code: "KeyA", altKey: true}),
+        new KeyboardEvent('keydown', { code: "KeyZ", altKey: true}),
```
the-t-in-rtf, line 56: That's funny.  Hopefully we won't have to get weird with the test cases as we keep adding keys.

src/App.js
```diff
@@ -652,6 +652,28 @@ export class App extends React.Component<AppProps, AppState> {
                                 return { announcementsEnabled: !(currentState.announcementsEnabled) };
                             });
                             break;
+                        case("addCommand"): {
+                            if (this.state.selectedAction) {
+                                const currentElement = document.activeElement;
+                                // $FlowFixMe: Not all elements have dataset property
+                                if (currentElement.dataset.controltype === 'programStep') {
```
the-t-in-rtf, line 659: It seems like this could avoid a lot of duplicated code by making a single check for the allowed control types and a ternary operator for the index variable.

chosww: Thanks for the suggestion, that's a good idea.

PR #218, opened 2021-07-12
C2LC-419: Focus and autoscroll to the command block added to the beginning and end

chosww:


the-t-in-rtf:
This works as advertised and, as Daniel and I discussed on the chats seems like it will should merge reasonably with #220.  Should be ready for further review.

chosww:
Merging this changes in another PR.

PR #226, opened 2021-07-23
C2LC-428: Add a keyboard shortcut to replace a step

chosww:
See [C2LC-428](https://issues.fluidproject.org/browse/C2LC-428) for details.


the-t-in-rtf:
I went through and did QA on this, I confirmed that the replace binding works in general.  I also confirmed that there are no issues with changing the selected command and hitting replace, i.e. that you never lose focus on the step to be replaced.  It works as expected on Chrome, FF, and Safari.

the-t-in-rtf:
The code changes seem in line with what was discussed on the chats, I'm fine with this moving on for further review.

the-t-in-rtf:
I verified the recent changes.  Very good catch about the action panel being open.  To test this, I opened the action panel, added commands at the beginning/end, changed the selected action using the keyboard, and then replaced the action I'd just added.  The animations also work consistently.

PR #229, opened 2021-07-29
C2LC-430: Add keyboard shortcut for the share button

chosww:
See [C2LC-430](https://issues.fluidproject.org/browse/C2LC-430) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -819,6 +819,9 @@ export class App extends React.Component<AppProps, AppState> {
                                 this.handleChangeCharacterPosition('turnRight');
                             }
                             break;
+                        case("shareEnvironment"):
+                            this.handleShareButtonShortcut();
+                            break;
                         default:
                             break;
                     }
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -1006,6 +1009,11 @@ export class App extends React.Component<AppProps, AppState> {
         this.setState({keyBindingsEnabled: keyBindingsEnabled});
     }
 
+    handleShareButtonShortcut = () => {
+        const currentUrl = document.location.href;
+        return navigator.clipboard.writeText(currentUrl);
```
the-t-in-rtf, line 1014: Does this mean there's no onscreen or announced feedback?   It seems like there should be.

the-t-in-rtf:
I left a comment here on the code but wasn't able to test the work in the time I have, as the preview deployment failed.

src/App.js
chosww: Didn't we say that the shortcut will just copy the url to the clipboard without the feedback message? And communicate more clearly with the name used in grid key for the switches, like "copy the current url"?

the-t-in-rtf: I still like letting people know that they've actually succeeded in hitting the key combination.  It doesn't have to be visual feedback, though.  Maybe this is a good use case for a preview announcement?

chosww: I agree that a preview(maybe feedback you mean?) announcement can make this clear.

the-t-in-rtf: Yes, I'm still getting them straight in my head, I guess in this case I did mean feedback.

the-t-in-rtf:
I performed QA against a copy of the branch checked out locally and confirmed that it works with FF, Chrome, and Safari.  I tested both keyboard input schemes.

PR #230, opened 2021-07-30
C2LC-431: Add Keyboard shortcut for expand and collapse add nodes

chosww:
See [C2LC-431](https://issues.fluidproject.org/browse/C2LC-431) for details.

### DIFF
--- a/src/KeyboardInputSchemes.js
+++ b/src/KeyboardInputSchemes.js
@@ -16,6 +16,7 @@ export type ActionName =
     | "addCommandToBeginning"
     | "addCommandToEnd"
     | "deleteCurrentStep"
+    | "expandAddNode"
     | "announceScene"
     | "decreaseProgramSpeed"
     | "increaseProgramSpeed"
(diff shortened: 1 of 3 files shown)


src/KeyboardInputSchemes.js
```diff
@@ -281,6 +282,10 @@ const VoiceOverInputScheme: KeyboardInputScheme = Object.assign({
     stopProgram: {
         keyDef: { code: "KeyS", key: "s", altKey: true},
         actionName: "stopProgram"
+    },
+    expandAddNode: {
```
the-t-in-rtf, line 286: For now this is fine, but as we pick up more bindings (like the help key) that are the same between all schemes, we should probably make and reuse a "common" binding definition, like we do with extended keyboard sequences.  I think a TODO here would be a good idea for now.

the-t-in-rtf:
I left a small comment on the code changes, but was not able to perform QA in the time I had, as the preview deployment failed.

src/KeyboardInputSchemes.js
chosww: Current keyboard shortcut combinations are likely to change, so I think it makes sense to wait until we talk about rework of keyboard shortcut combinations.

the-t-in-rtf:
I performed QA on this with a local copy of the branch. It works as expected on all browsers, and I confirmed that the change is equivalent to hitting the toggle, i.e.:

1. Dragging and dropping with contracted nodes works as expected if you contract the nodes using the hotkey.
2. Tab navigation works as expected if you contract the nodes using the hotkey.

PR #231, opened 2021-07-30
C2LC-433: Add keyboard shortcut for pen down toggle

chosww:
See [C2LC-433](https://issues.fluidproject.org/browse/C2LC-433) for details.


the-t-in-rtf:
The code changes look reasonable, but the preview deployment failed, so I didn't perform any QA.

the-t-in-rtf:
I performed QA on a local checkout of the branch. It works as expected on all browsers.  I specifically tested using the key binding while the program is playing, which works the same as if you'd toggled the pen up/down using a keyboard or mouse.  This is ready for further review as far as I can see.

PR #235, opened 2021-08-23
C2LC-437: Change ThemeSelector to a modal in the new design

chosww:
See [C2LC-437](https://issues.fluidproject.org/browse/C2LC-437) for details.


the-t-in-rtf:
I checked out the preview site in Chrome, Firefox, and Safari.  It works as expected, the theme changes in the background as you select a new theme, but reverts as expected if you hit cancel.   I see a note on the figma mockups about adding a preview screen, not sure if that's still something that needs to be done.

In general it looks good, the colours and the hover/focus of the button are spot on.  The vertical spacing of the modal itself seems to need something.  Here's the mockup:

Screenshot 2021-08-24 at 14 21 15

And here's the same at the default sizing in Firefox:

Screenshot 2021-08-24 at 14 21 26

Seems to me like the vertical padding for individual items is too tight,. The header could use more vertical padding  as well.  The footer seems to have a bit too much vertical padding.

the-t-in-rtf:
Looking at the screen captures together, it also seems like the contrast icon and the radio buttons are not left indented quite far enough.

the-t-in-rtf:
Checked a range of themes and only found one issue.  We might want to file another small ticket for this, I notice that both the keyboard and theme toggles "jump" in high contrast mode.

the-t-in-rtf:
The keyboard support in general is good.  I do kind of think that hitting enter on an option should be equivalent to clicking "done".  At the moment you have to tab to "done" and then hit "enter".  By comparison, hitting escape immediately cancels, which is what I'd expect from "enter".  We should talk it through in the meeting.

the-t-in-rtf:
> I do kind of think that hitting "enter" on an option should be equivalent to clicking "done".

I see that the behaviour is the same for the keyboard modal, but it's a good thing to discuss as we work to standardise our modals in general.

the-t-in-rtf:
A couple more things for today's meeting.  First, we should discuss how this work relates to refactoring modals in general.  Second, it seems like there are a bunch of older issues in the todo list that probably could move to "in progress" and be assigned to Daniel, as his work here will resolve them.

the-t-in-rtf:
Did some more testing this morning following up on yesterday's meeting.  I discovered a small polish issue at the default resolution in FF (all themes):

Screenshot 2021-08-25 at 09 43 18
Screenshot 2021-08-25 at 09 43 12

The focus border is clipped on the bottom and sides.

I also noticed that the toggle icons don't scale correctly in Firefox (all themes):

Screenshot 2021-08-25 at 09 42 45

The scene refresh icon does scale, so I assume it's not just a feature of FF.

src/Themes.scss
```diff
@@ -1400,7 +1398,7 @@ body.contrast-theme {
                 fill: white;
             }
         }
-    
+
         &:focus {
             box-shadow: 0 0 0 $c2lc-focus-indicator-width white;
             border: 2px solid black;
```
the-t-in-rtf, line 1404: In my testing, this is the line that's making the icon jump.

the-t-in-rtf:
The new modal has the same problem described in [C2LC-439](https://issues.fluidproject.org/browse/C2LC-439), i.e. shift+tab never wraps around to the last focusable element.

chosww:
> Did some more testing this morning following up on yesterday's meeting. I discovered a small polish issue at the default resolution in FF (all themes):
> 
> Screenshot 2021-08-25 at 09 43 18
> Screenshot 2021-08-25 at 09 43 12
> 
> The focus border is clipped on the bottom and right side.
> 
> I also noticed that the icons to open the keyboard and theme modals don't scale correctly in Firefox (all themes):
> 
> Screenshot 2021-08-25 at 09 42 45
> 
> The scene refresh icon does scale, so I assume it's not just a feature of FF.

Weirdly I can't reproduce the issues on my Mac and PC, what OS, screen scale level, and the browser zoom level did you use to get these?

the-t-in-rtf:
> Weirdly I can't reproduce the issues on my Mac and PC, what OS, screen scale level, and the browser zoom level did you use to get these?

Firefox on OS X, I'll make a screen capture for you.

the-t-in-rtf:
FF icon zoom

src/ThemeSelector.js
```diff
@@ -1,42 +1,153 @@
...
-                    <Dropdown.Item eventKey='gray'>{this.props.intl.formatMessage({id:'ThemeSelector.theme.gray'})}</Dropdown.Item>
-                }
-                { // $FlowFixMe
-                    <Dropdown.Item eventKey='contrast'>{this.props.intl.formatMessage({id:'ThemeSelector.theme.contrast'})}</Dropdown.Item>
-                }
-            </DropdownButton>
+            <ModalWithFooter
+                show={this.props.show}
+                focusOnOpenSelector={`.ThemeSelector__option.${this.props.currentTheme}`}
+                focusOnCloseSelector='.IconButton.themeSelector'
```
the-t-in-rtf, line 129: The close selector isn't working at the moment, that button has `class="IconButton App__header-themeSelectorIcon"`.   So you'd need `.IconButton.App__header-themeSelectorIcon` instead.

the-t-in-rtf:
I found a small issue with the focus on close, other than that this seems good.

src/ThemeSelector.js
chosww: Thanks @the-t-in-rtf, nice catch 👍

src/App.scss
```diff
@@ -36,30 +36,16 @@ body {
...
-    &:focus {
-        box-shadow: 0 0 0 $c2lc-focus-indicator-width black;
-        outline: none;
-    }
-}
-
-.App__header-keyboardMenuIcon--disabled {
-    background-color: #B9BEC6;
+.App__header-menu {
+    // Uncomment background-color and border ladius later when we change all the setting
```
sbates-idrc, line 40: Typo: "radius"

sbates-idrc:
Hi Daniel, the PR looks great.

I've created a C2LC-437 branch with some changes: renaming the "Done" button to "Save", changing the background colour of the themes list, and some code changes to try to make some things a little clearer.

https://github.com/sbates-idrc/c2lc-coding-environment/tree/C2LC-437

Could you please check my changes over? If they look good to you, I will then merge. Thanks very much.

I also noticed, but didn't change yet: there are a number of messages named "ThemeSelector.theme*" Which I think can be removed now?

PR #238, opened 2021-08-30
C2LC-436: Refactor the code to handle user actions for changing the program sequence

sbates-idrc:
This PR refactors the code to handle user actions for changing the program sequence (currently the interactions in the `ProgramBlockEditor` and keyboard shortcut handlers in `App`).

It introduces a new class `ProgramChangeController` which is now responsible for making changes to the `App` `state.programSequence` and coordinating any user interface activities associated with the change, such as making announcements, setting focus, or setting up animations.

Handlers for user actions for changing the program sequence now delegate to the `ProgramChangeController`, to ensure consistent user experience across different modes of access.

The details for a particular user action (index, selectedAction, step to delete, and such) are captured at the handler and then provided to the `ProgramChangeController`. With this approach, the `ProgramChangeController` can do checks, such as checking that the step to delete hasn't been replaced by another change to the program that was applied before the deletion was processed (in the case that React has batched state updates).

### DIFF
--- a/.flowcoverage-tests-threshold
+++ b/.flowcoverage-tests-threshold
@@ -1,5 +1,8 @@
 {
   "concurrentFiles": 1,
+  "globExcludePatterns": [
+    "src/ProgramChangeController.test.js"
+  ],
   "globIncludePatterns": [
     "src/**/*.test.js"
   ],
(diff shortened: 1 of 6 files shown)


chosww:
Thanks @sbates-idrc for the refactoring work, the changes look good and reflects the details you have written in the description of this PR.

.flowcoverage-tests-threshold
```diff
@@ -1,5 +1,8 @@
 {
   "concurrentFiles": 1,
+  "globExcludePatterns": [
+    "src/ProgramChangeController.test.js"
```
the-t-in-rtf, line 4: Were there issues typing material in this file?  Do we have an option to exclude those parts, perhaps using `$FlowIgnore`?

the-t-in-rtf:
I went through QA, focusing particularly on state changes both when stopped and playing.  I didn't see any regressions on previous issues, such as making changes with the keyboard when the program is running, leaving the action panel open while running.  Everything worked as expected.

I also looked at the code changes (and the discussion around them on the chats).  Other than resolving the conflict, I only had one small question about flow type coverage (see above).

PR #240, opened 2021-09-01
C2LC-448: Added ctrl+shift key bindings.

the-t-in-rtf:
See [C2LC-448](https://issues.fluidproject.org/browse/C2LC-448) for details.


chosww:
The changes look good, and works well on both Windows and MacOS.

the-t-in-rtf:
I merged with develop 1.2 and fixed a linting error introduced by the merge, this one is clean again.

the-t-in-rtf:
The "escape" tests for the keyboard handler were trying to toggle the audio settings, which didn't work following the audio settings refactor.  I changed that to use a different setting, as we were just testing "escape" functionality there.  I will also review to make sure we didn't lose any key bindings during the merge.

the-t-in-rtf:
Seems fine, although we only have a shortcut to toggle announcements at the moment.

PR #241, opened 2021-09-01
C2LC-446: Added persistence for keyboard settings.

the-t-in-rtf:
See [C2LC-446](https://issues.fluidproject.org/browse/C2LC-446) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1391,7 +1391,7 @@ export class App extends React.Component<AppProps, AppState> {
         const localKeyBindingsEnabled = window.localStorage.getItem('c2lc-keyBindingsEnabled');
         const localKeyboardInputSchemeName = window.localStorage.getItem('c2lc-keyboardInputSchemeName');
 
-        if (localKeyBindingsEnabled !== null) {
+        if (localKeyBindingsEnabled != null) {
             try {
                 this.setState({
                     keyBindingsEnabled: JSON.parse(localKeyBindingsEnabled)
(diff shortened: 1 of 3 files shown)


chosww:
Thanks @the-t-in-rtf, the changes look good. Have we decide not to include the keyboard shortcut stuffs as URL parameters? If so, I think this work is ready for @sbates-idrc to review.

the-t-in-rtf:
In meetings I had proposed leaving it out of the URL, as we don't want to override someone's keyboard settings when they make use of a link another user has shared.

src/App.js
```diff
@@ -1386,6 +1387,30 @@ export class App extends React.Component<AppProps, AppState> {
             });
         }
 
+        // Keyboard settings are read from local storage whether or not we have URL content.
+        const localKeyBindingsEnabled = window.localStorage.getItem('c2lc-keyBindingsEnabled');
+        const localKeyboardInputSchemeName = window.localStorage.getItem('c2lc-keyboardInputSchemeName');
+
+        if (localKeyBindingsEnabled !== null) {
```
sbates-idrc, line 1394: I think in this case that it's a little safer to use the less strict '!=' as it will catch `undefined`:

```
> undefined !== null
true
> undefined != null
false
```

src/App.js
```diff
@@ -1386,6 +1387,30 @@ export class App extends React.Component<AppProps, AppState> {
...
+            }
+            catch(err) {
+                /* eslint-disable no-console */
+                console.log(`Error parsing key bindings toggle: ${localKeyBindingsEnabled}`);
+                console.log(err.toString());
+                /* eslint-enable no-console */
+            }
+        }
+
+        if (localKeyboardInputSchemeName !== null) {
```
sbates-idrc, line 1408: As above. I think '!=' is safer here.

src/App.js
the-t-in-rtf: Good point, and fixed.

src/App.js
the-t-in-rtf: Fixed as well.

PR #242, opened 2021-09-01
C2LC-429: Added shortcut to delete last program step.

the-t-in-rtf:
See [C2LC-429](https://issues.fluidproject.org/browse/C2LC-429) for details.


chosww:
Thanks @the-t-in-rtf, I read the chat you had with Simon in Element, and the changes you made looks good based on the chat. Test case looks good as well.

PR #243, opened 2021-09-02
C2LC-447: Reduce the number of grid squares (increase size)

chosww:
See [C2LC-447](https://issues.fluidproject.org/browse/C2LC-447) for details.

### DIFF
--- a/src/Scene.js
+++ b/src/Scene.js
@@ -52,8 +52,7 @@ class Scene extends React.Component<SceneProps, {}> {
                     key={`grid-cell-label-${i}`}
                     dominantBaseline='middle'
                     x={-0.5}
-                    // Center each gridcell with height of 8.25
-                    y={8.25*i - 4.125}>
+                    y={i*10 - 5}>
                     {i}
                 </text>
             )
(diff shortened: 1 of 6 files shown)


the-t-in-rtf:
I tested a bunch of things based on previous work, all of which worked fine:

1. Side wall collisions were handled correctly for all four walls.
2. Corner collisions were handled correctly.
3. Both were handled correctly regardless of the distance from the boundary (forward3 when one square away, forward3 when two squares away, et cetera).
4. When zoomed in enough that scrolling happens, auto-scrolling works as expected in all three browsers.
5. The sounds pan correctly.

The only thing I found was this zooming issue at 240% in Firefox:

Zooming Issues Firefox

Based on our previous discussions, I checked, and this only happens when "zoom text only" is checked.

That's it for QA, I'll comment on the code shortly.

src/Scene.js
```diff
@@ -52,8 +52,7 @@ class Scene extends React.Component<SceneProps, {}> {
                     key={`grid-cell-label-${i}`}
                     dominantBaseline='middle'
                     x={-0.5}
-                    // Center each gridcell with height of 8.25
-                    y={8.25*i - 4.125}>
+                    y={i*10 - 5}>
```
the-t-in-rtf, line 55: I think it would be nice to update and keep the comment.

the-t-in-rtf:
I left a very minor comment on the code, otherwise looks good.

PR #244, opened 2021-09-08
C2LC-440: Create modal for changing the background of the coding environment

chosww:
See [C2LC-440](https://issues.fluidproject.org/browse/C2LC-440) for details.


sbates-idrc:
Thanks Daniel. It's so great to have the new backgrounds!

sbates-idrc:
As we discussed, I have started some refactoring work to collect the world information together into a single service. So rather than the world information (which background SVG to use, which character SVG to use, and such) being distributed throughout the application, there will be a single place in the application that gathers that information together. This will make it easier in the future to add or remove or edit worlds if there is a single place to look. And if we want to add lazy-loading of world assets in the future, introducing a world service now will make that easier as the application code is isolated from where the assets come from: they just ask the service.

PR #245, opened 2021-09-15
C2LC-458: Disable key bindings by default.

the-t-in-rtf:
See [C2LC-458](https://issues.fluidproject.org/browse/C2LC-458) for details.


chosww:
The changes look good, and I confirmed that the preview site has keyboard shortcut disabled by default in different browsers.

PR #246, opened 2021-09-17
C2LC-451: Set character and starting and reset position based on the background

chosww:
See [C2LC-451](https://issues.fluidproject.org/browse/C2LC-451) for details.


src/CharacterPositionController.test.js
```diff
@@ -199,15 +200,32 @@ describe('Using change character position by column/row labels', () => {
...
+        wrapper.setProps({world: 'DeepOcean', theme: 'light'});
+        expect(getCharacterIcon(wrapper).get(0).type.render().props.children).toBe('Submarine.svg');
+        wrapper.setProps({theme: 'gray'});
+        expect(getCharacterIcon(wrapper).get(0).type.render().props.children).toBe('Submarine-gray.svg');
+        wrapper.setProps({theme: 'contrast'});
+        expect(getCharacterIcon(wrapper).get(0).type.render().props.children).toBe('Submarine-contrast.svg');
     });
     test('Character icon gets transform value to match the character in the scene', () => {
-        expect.assertions(3);
+        //expect.assertions(3);
```
sbates-idrc, line 228: Please uncomment this line.

src/Worlds.test.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+import { getWorldCharacter, getWorldProperties, isWorldName } from './Worlds';
+
+import { ReactComponent as Submarine } from './svg/Submarine.svg';
+import { ReactComponent as SubmarineGray } from './svg/Submarine-gray.svg';
+import { ReactComponent as SubmarineContrast } from './svg/Submarine-contrast.svg';
+
+test('isWorldName', () => {
```
the-t-in-rtf, line 9: I realise some of this is not new to this pull, but it's the first time I'm looking through the `isWorldName` stuff in this depth.  I wanted to ask why we have `isWorldName` rather than just having a longer list of switches in the case statement (as we do with the themes).

It actually seems a line or two longer than just making a longer case statement, and adds a function to Worlds.js that seems unnecessary.

the-t-in-rtf:
I tested this in some depth.  The persistence and reset seem to work just fine for all worlds.   I had one code question, otherwise this seems ready for review.

I did find some small issues with individual world and theme combinations, I will file those as separate issues.

src/Worlds.test.js
sbates-idrc: The `isWorldName` function is used by `getWorldFromString` in `Utils.js`: https://github.com/chosww/c2lc-coding-environment/blob/7c5c7b0ea04dbc19c7f526c4926d328e427d2b71/src/Utils.js#L48-L64

The intention behind this implementation is to minimize the number of places in the code that know what worlds are available. We are planning on adding to the available worlds in the future and if we can locate knowledge of what worlds are available in `Worlds.js`, then we minimize the number of places in the code that we will need to remember to update when new worlds are added.

In this particular case, if the `getWorldFromString` function contained a list of the available worlds, and we didn't remember to update that list when we added new worlds, then it would introduce a potentially hard to find bug. The new worlds would mostly work just fine. And the bug would only appear when we tried to load a URL that used one of the new worlds, where the default world would be used instead.

PR #247, opened 2021-09-17
C2LC-454: Add Special Squares to the Character Announcement

chosww:
See [C2LC-454](https://issues.fluidproject.org/browse/C2LC-454) for details.

Note that this work has merged C2LC-451, as some of the changes in this branch has moved to C2LC-451.


sbates-idrc:
This PR includes the changes from C2LC-451 which hasn't been merged to `develop-0.9` yet (as of the time of writing this). Which makes this PR look larger than it is.

src/App.js
```diff
@@ -111,7 +111,7 @@ export class App extends React.Component<AppProps, AppState> {
     interpreter: Interpreter;
     audioManager: AudioManager;
     focusTrapManager: FocusTrapManager;
-    startingCharacterState: CharacterState;
+    startingCharacterStates: any;
```
the-t-in-rtf, line 114: It's trivial to make a type for all the supported character states, and I think we should unless we have a reason not to.

src/Worlds.js
```diff
@@ -26,58 +36,95 @@ import { ReactComponent as SpaceContrast } from './svg/Space-contrast.svg';
...
+    thumbnailContrast: React.ComponentType<{}>,
+    character: React.ComponentType<{}>
 };
 
 const worlds = {
     'DeepOcean': {
         background: DeepOcean,
         backgroundGray: DeepOceanGray,
         backgroundContrast: DeepOceanContrast,
+        backgroundInfo: [
```
the-t-in-rtf, line 58: I understand that a 2-dimensional array is a simple way of expressing this, but I think there are other ways to encode this that would require less `null` values, and that would be more flexible if we decide to change the scene size again.

For example:

```
{
  1: { 8: "DeepOcean.fishGroup", 9: "DeepOcean.fishGroup" },
  2: { 8: "DeepOcean.fishGroup", 9: "DeepOcean.fishGroup" },
  3: { 1: "DeepOcean.babyJellyfish", 2: "DeepOcean.babyJellyfish", 6: "DeepOcean.shark", 7: "DeepOcean.shark", 10: "DeepOcean.shark", 11: "DeepOcean.coral" }
  // etc.
}
```

You'd need a slightly different check to see if it's set, something like:

```
let cellContent = backgroundInfo[row] && backgroundInfo[row][col];
if (cellContent) { 
  // Do stuff...
}
```

But you can omit all empty cells and even whole empty rows with that kind of approach, and still  check for content in the cell easily.

src/Worlds.js
```diff
@@ -26,58 +36,95 @@ import { ReactComponent as SpaceContrast } from './svg/Space-contrast.svg';
...
+        character: Submarine,
+        characterGray: SubmarineGray,
+        characterContrast: SubmarineContrast
     },
     'Jungle': {
         background: Jungle,
         backgroundGray: JungleGray,
         backgroundContrast: JungleContrast,
+        backgroundInfo: [
+            [null, null, null, null, null, null, 'Jungle.giraffe', null, null, 'Jungle.lion', null, null],
```
the-t-in-rtf, line 80: I also think we could do more with type checking here, whether or not we change the structure.  One approach would be to have a type for the items in each group, and to OR those together if we want a set of all supported items.

I think it's probably a good idea to decouple the item name from message keys as well, as we might want to do things other than just announce them.  So, you might choose to announce `"{World}.{cellContents}"`, but I might want to check `cellContents` separately and play a sound or animation for "giraffe" or "lion".  It makes more sense to keep the message key scoping out in that case.

It also makes it easier to share objects between worlds, for example, if we add goals or other things that are the same in multiple worlds.

src/Worlds.js
```diff
@@ -90,3 +137,22 @@ export function isWorldName(str: ?string): boolean {
 export function getWorldProperties(world: WorldName): WorldProperties {
     return worlds[world];
 }
+
+export function getBackgroundInfo(world: WorldName, yPos: number, xPos: number): ?string {
+    const worldBackgroundInfo = worlds[world].backgroundInfo;
+    if (worldBackgroundInfo) {
```
the-t-in-rtf, line 143: I was mentioning making an alternate structure, if you want to be able to omit `backgroundInfo` altogether, then that check would be more like `backgroundInfo && backgroundInfo[yPos] && backgroundInfo[xPos]`.

src/CharacterAriaLive.js
```diff
@@ -16,50 +17,46 @@ type CharacterAriaLiveProps = {
...
-        ariaLiveRegion.innerText=this.props.intl.formatMessage(
-            {id:'CharacterAriaLive.positionAriaLabel'},
-            {
-                character,
-                xPos,
-                yPos,
-                direction
-            }
-        );
+        const backgroundInfo = getBackgroundInfo(this.props.world, xPos - 1, yPos - 1);
```
the-t-in-rtf, line 36: If we went with my object based strategy, we could also avoid "off by one" thinking like this, and just let row and cell one mean the same thing for objects in the scene as they do for the character position.

src/Worlds.js
```diff
@@ -90,3 +137,22 @@ export function isWorldName(str: ?string): boolean {
...
+        return worldBackgroundInfo[yPos][xPos];
```
the-t-in-rtf, line 144: I can actually get the environment to crash, which is a sign to me that there are cases in which `worldBackgroundInfo` is set but not all rows/columns are.  Try moving one square from H7 to I7 in the Space world.

the-t-in-rtf: Looks like anything higher than H causes a crash.

the-t-in-rtf:
I think your rows/columns are transposed.  In the Jungle theme, the baby alligator announces on F5 and F6, when it should announce on E6 and F6.  The giraffe is announced on A7.  This would also explain the crashes when moving to column I, which is the first beyond the number of rows we have.  I will discontinue further testing for now, as I can only test part of the scene at the moment.

On a side note, if you could merge with upstream/develop-0.9, it would be helpful in testing, as I'd like it to remember to use the VoiceOver key bindings.

src/CharacterAriaLive.js
the-t-in-rtf: This is also exactly where you transposed the coordinates.  It should be `yPos -1` first.

the-t-in-rtf:
I was just thinking about the transposition problems.  Seems like an integration test where you jump to a cell, trigger, and check the announcement might be good.

src/Worlds.js
chosww: Crash was caused by calling the method in CharacterAriaLive.js and Scene.js with wrong parameter order

PR #248, opened 2021-09-17
C2LC-457: Set world when one of the thumbnail images from the world selector is clicked

chosww:
See [C2LC-457](https://issues.fluidproject.org/browse/C2LC-457) for details.


the-t-in-rtf:
This works as advertised, but I'm a little concerned about the user experience:

1. You can't hit enter to select a world when its icon has focus (you can hit escape to close the dialog, though).
2. The navigation switches from tabs to arrows when you hit the world icons.

One way to address both would be to get rid of the checkboxes and just make the icons buttons.

PR #251, opened 2021-09-21
C2LC-460: Update background of character icons 

chosww:
See [C2LC-460](https://issues.fluidproject.org/browse/C2LC-460) for details


the-t-in-rtf:
The code changes here seem fine, and the contrast is better, especially for the greyscale theme.  Ready for further review.

PR #253, opened 2021-09-22
C2LC-465: Update drawing path in high contrast theme 

chosww:
See [C2LC-465](https://issues.fluidproject.org/browse/C2LC-465) for details.

### DIFF
--- a/src/Worlds.scss
+++ b/src/Worlds.scss
@@ -37,7 +37,7 @@ body.contrast-theme {
     }
 
     .Scene__path-line--DeepOcean {
-        stroke: #FFF500;
+        stroke: #FF0000;
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/Worlds.scss
```diff
@@ -176,6 +176,6 @@ body.contrast-theme {
     }
 
     .Scene__path-line--Space {
-        stroke: #FFF500;
```
the-t-in-rtf, line 179: Since it's now the same colour for all worlds, it seems simpler just to have one rule for `.Scene__path-line`

the-t-in-rtf:
The changes look good, I checked all worlds and looked for issues with legibility, layering.

I had a small comment on consolidating the rules now that all worlds use the same colour, otherwise this is ready for review.

src/Worlds.scss
chosww: With same reasoning as the comment I left in PR-469, this is some duplication we are expecting.

PR #255, opened 2021-09-22
C2LC-467: Shift the scene background image to the left

chosww:
See [C2LC-467](https://issues.fluidproject.org/browse/C2LC-467) for details.


the-t-in-rtf:
For a fullscreen window, there is no longer overlap in the bottom right corner between the buttons and background elements.  It's still possible to overlap if your browser window is narrow, for example the buttons end up overlapping with the treasure in the Deep Ocean world.  However, in portrait mode on an iPad there's no overlap, so this seems like a good balance to me.

The code changes are fine.  I'd say this is ready for further review, although we might want to discuss the issues with very narrow windows.

PR #257, opened 2021-09-22
C2LC-469: Update grid line colour in the high contrast theme

chosww:
See [C2LC-469](https://issues.fluidproject.org/browse/C2LC-469) for details.

### DIFF
--- a/src/Worlds.scss
+++ b/src/Worlds.scss
@@ -33,7 +33,7 @@ body.contrast-theme {
     }
 
     .Scene__grid-line--DeepOcean {
-        stroke: #FFFFFF;
+        stroke: #0078FF;
     }
 
     .Scene__path-line--DeepOcean {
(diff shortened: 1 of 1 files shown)


src/Worlds.scss
```diff
@@ -172,7 +172,7 @@ body.contrast-theme {
     }
 
     .Scene__grid-line--Space {
-        stroke: #FFFFFF;
+        stroke: #0078FF;
```
the-t-in-rtf, line 175: As with the drawing colour, if they're the same, I'd argue we should just have a single rule for `.Scene__grid-line`.

the-t-in-rtf:
Although I have to imagine how it'll look with the updated drawing colour, the grid line colours are fine.  I left a small comment about consolidating the world styles a bit.

src/Worlds.scss
chosww: .Scene__grid-line will only have common properties, and we would specify colours in different worlds with --{WorldName} class name. This may introduce some duplication in the codebase, but would be easier to understand the codebase.

PR #258, opened 2021-09-22
C2LC-461: Adjusted spaceship icons.

the-t-in-rtf:
See [C2LC-461](https://issues.fluidproject.org/browse/C2LC-461) for details.


the-t-in-rtf:
I added a circle as a guide and found that resizing to 90px wide was enough to fix the problems.  I used Inkscape, and tried to clean up all its cruft, but the stroke width was scaled proportionately.  Not sure if we want to tweak that.

I did briefly try adjusting to see if I could centre the design better so that the nose would point into the corners better, but couldn't easily manage it.  That'll have to wait for a new ticket or design from Sepi.

chosww:
Instead of changing the svg contents of the space ship, I suggest reducing the width of the icon that gets rendered. Line 367 of Scene.js, we are passing value 0.9 as width, and reducing it to 0.8 seems to fit the spaceship character within the grid.

the-t-in-rtf:
I made the suggested change.  That also changes the size of the other icons, but not in a bad way.  The jeep was a bit tight before.

chosww:
A test is failing because of the changed width value. This failure is a little unexpected, and I see that we have a TODO comment on line 81 of Scene.test.js. It would be great if you can think of a better way to make that update in the TODO, or just update characterWidth in that function to 0.8 for this release.

the-t-in-rtf:
Done.  I couldn't think of a good way to detect the number without making it brittle in other ways, so I just changed it (back) to 0.8 (the comment still had the old value).

chosww:
Thanks Tony, let's revisit about the TODO note.

sbates-idrc:
The spaceship character has been redesigned since this issue was created. Closing.

PR #260, opened 2021-09-22
C2LC-459: Fixed blurry world picker radio buttons on iOS.

the-t-in-rtf:
See [C2LC-459](https://issues.fluidproject.org/browse/C2LC-459) for details.


chosww:
As we decided to remove the custom style and use the default checkbox style, the change looks good.

PR #261, opened 2021-09-22
C2LC-472: Fixed padding of share confirmation dialog.

the-t-in-rtf:
See [C2LC-472](https://issues.fluidproject.org/browse/C2LC-472) for details.

### DIFF
--- a/src/ShareCompleteModal.css
+++ b/src/ShareCompleteModal.css
@@ -3,6 +3,9 @@
 }
 
 .ShareCompleteModal__content {
-    border-radius: 0.25rem;
     background-color: yellow;
+}
+
+.ShareCompleteModal__content.modal-body {
+    padding: 1rem;
 }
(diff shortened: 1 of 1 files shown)


chosww:
The fix causes double line message, which looks little weird with so much padding to the right side. I suggest adding width: max-content to the message's container (.modal-content).

Screen Shot 2021-09-22 at 1 19 01 PM
au

the-t-in-rtf:
I agree, and I've updated to use a slightly larger max-width.  I also centred the text.

chosww:
I am thinking, instead of using Modal.Body component for the text, we can just use div to wrap up the text. By using Modal.Body, React Bootstrap is including some class names to the text container those are not in our source code, and to change their default style, we need to use unwanted chaining in the style sheet. 

The change looks good, but we can have less and cleaner css rules by changing the wrapper to just div. I will assign this to Simon to see what they think about it as well.

PR #262, opened 2021-09-22
C2LC-471: Tamed keyboard input modal styles breaking other modals.

the-t-in-rtf:
See [C2LC-471](https://issues.fluidproject.org/browse/C2LC-471) for details.

### DIFF
--- a/src/KeyboardInputModal.scss
+++ b/src/KeyboardInputModal.scss
@@ -2,6 +2,13 @@
 
 .KeyboardInputModal {
     max-width: 75%;
+}
+
+.KeyboardInputModal__content {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    margin-top: 1rem;
 
     .modal {
         border-radius: 0.25rem;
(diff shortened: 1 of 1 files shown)


chosww:
The changes fix the issue, but I suggest changing css class name for KeyboardInputModal to be more specific to itself, something like KeyboardInputModal__content or something.

the-t-in-rtf:
Sounds good, I moved the new directives under that class name instead.

src/KeyboardInputModal.scss
```diff
@@ -18,6 +9,15 @@
     flex-direction: column;
     justify-content: center;
     margin-top: 1rem;
+
+    .modal {
```
chosww, line 13: .modal is not a child of .KeyboardInputModal__content, so this rule is not applying anything, we can remove it.

src/KeyboardInputModal.scss
```diff
@@ -18,6 +9,15 @@
...
+        border-radius: 0.25rem;
+    }
+    
+    .modal-body {
```
chosww, line 17: same for the .modal-body, but we need padding:0 directly applied to .KeyboardInputModal__content, so updated KeyboardInputModal__content would be:

```
.keyboardInputModal__content {
display: flex;
flex-direction: column;
justify-content: center;
margin-top: 1rem;
padding: 0;
}
```

src/KeyboardInputModal.scss
the-t-in-rtf: Good catch, and done.

PR #263, opened 2021-09-22
C2LC-473: Removed world thumbnail from right-hand panel.

the-t-in-rtf:
See [C2LC-473](https://issues.fluidproject.org/browse/C2LC-473) for details.


chosww:
Removing the rendering part looks good, but could you also remove css rules for the world thumbnail in App.scss please?

the-t-in-rtf:
Good catch, and done.

PR #264, opened 2021-09-22
C2LC-475: Comment out css animations 

chosww:
See [C2LC-475](https://issues.fluidproject.org/browse/C2LC-475) for details.

Note that I removed --updated className in Javascript files, and I removed --paused style in SCSS files, as some test cases, we are checking for --paused className.


sbates-idrc:
With this PR there is no longer any indication of the step that the program is paused on. We discussed including some styling for the paused step, but without animation. I'm also re-thinking commenting out vs removing the code and now I'm thinking that it would be better to just remove the animation code. We can find it again in the git history.

chosww:
It was a software bug, after updating iOS version, it got fixed.

PR #266, opened 2021-09-23
C2LC-481: Updated keyboard/world picker modals to use 'save'.

the-t-in-rtf:
See [C2LC-481](https://issues.fluidproject.org/browse/C2LC-481) for details.


chosww:
The changes look good.

src/ModalFooter.js
```diff
@@ -21,7 +21,7 @@ class ModalFooter extends React.Component<ModalFooterProps, {}> {
                     </Button>
                 }
                 <Button className="ModalFooter__doneButton" onClick={this.props.onClickDone}>
```
sbates-idrc, line 23: Please also rename the callback property to `onClickSave`, and any related handlers to `handleClickSave` or such.

sbates-idrc: Actually. I'm having some different thoughts about this now.

sbates-idrc: I think we want the `ModalFooter` to be more flexible than it currently is. Right now we have:

* Always a 'Done' button, with callback `onClickDone` which this PR gives the label "Save"
* And an optional 'Cancel' button, with callback `onClickCancel`

Looking at the Figma designs, I think we will need a few different variations:

* Cancel, Save (Available Actions, Background, Keyboard Shortcuts, Sound Options, Themes)
* Cancel (Share Link, though maybe this should be "Done" rather than "Cancel"?)
* Cancel, Delete (Delete Program)
* Cancel, Connect (Robot Connections, though I think that we cannot make this UI due to limitations in the functionality of the web Bluetooth API)

So mostly we have (Cancel, Save), but we also have (Cancel, Delete) and possibly Cancel only (though that might be better Done only).

sbates-idrc: A possibility would be to change the `ModalFooter` properties to take a list of (label, callback) pairs. That way the footer could display any number of buttons, with different semantics. Something like:

```
<ModalFooter
    buttons={[
        {
            label: this.props.intl.formatMessage({ id: 'CancelButton' }),
            callback: this.handleClickCancel
        },
        {
            label: this.props.intl.formatMessage({ id: 'SaveButton' }),
            callback: this.handleClickSave
        }
    ]}
/>
```

the-t-in-rtf:
As discussed, this is superseded by the work on the new modal.

PR #270, opened 2021-09-23
C2LC-480: Update text colours in different UI components

chosww:
See [C2LC-480](https://issues.fluidproject.org/browse/C2LC-480) for details.


the-t-in-rtf:
These are hard to eyeball, so I used the FF accessibility checker.  It doesn't flag the other instances, but definitely is happier with the text of the "Theme" button with the changes.  Seems ready for further review.

PR #271, opened 2021-09-23
C2LC-486: Fix broken keyboard shortcuts

chosww:
See [C2LC-486](https://issues.fluidproject.org/browse/C2LC-486) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -824,7 +824,7 @@ export class App extends React.Component<AppProps, AppState> {
                             focusOnFirstElementWithClass("command-block");
                             break;
                         case("focusAppHeader"):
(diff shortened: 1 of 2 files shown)


the-t-in-rtf:
The strategy is a little brittle, I'm not surprised we ended up here.  It seems like we should add focus tests if we can do so easily.  Otherwise we risk regressions on this as we refactor class names, et cetera.

chosww:
I think scope of this task would be simply to fix the broken keyboard shortcuts, and I introduced this className specific for keyboard shortcuts to indicate that the elements have focus keyboard shortcuts; giving specific className like this is used in keyboard trap items as well. We need another task later to update all elements those have keyboard shortcut to has more specific className, and maybe include some test cases to confirm the behaviour.

PR #272, opened 2021-09-24
C2LC-478: Update thumbnail images for the world

chosww:
See [C2LC-478](https://issues.fluidproject.org/browse/C2LC-478) for details.


the-t-in-rtf:
This is much improved, all corners match now.  I'm a little concerned about the mismatch between the borders of the sketchpad and jungle icons in the default/light/dark themes.  The jungle icon has a dark black border, while the sketchpad has a grey border.   If you look at the high contrast theme by comparison, three of the world icons have borders, and they match.

chosww:
I've updated with the new thumbnail images. With the changes, there's a little weird light boarder on thumbnail preview image of Sketchpad world in the high contrast theme, but we are going to remove the thumbnail image in the other PR, so it won't be an issue.

PR #273, opened 2021-09-24
C2LC-479: Remove disabled property from IconButton

chosww:
See [C2LC-479](https://issues.fluidproject.org/browse/C2LC-479) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1138,7 +1138,6 @@ export class App extends React.Component<AppProps, AppState> {
                                 </a>
                             </h1>
                             <IconButton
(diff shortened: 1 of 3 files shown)


chosww:
I initially thought we don't need disabled property for IconButton component, but we should update any circular buttons with it. Currently, we have 2 more icon buttons those are not implemented using IconButton, and they are RefreshButton and DeleteAllButton. When we change them to use IconButton, we need disabled property, and the disabled property name can stay and used for actual disabled state. So we can remove TODO when we are working on the task.

the-t-in-rtf:
The code changes look good and work as expected in the first three themes.

PR #275, opened 2021-09-29
C2LC-489: Make it clear that share confirmation is a dialog.

the-t-in-rtf:
See [C2LC-489](https://issues.fluidproject.org/browse/C2LC-489) for details.


the-t-in-rtf:
@chosww , as discussed in the chats, I did try the `aria-labelledby` approach, but it didn't change what was announced at least in VoiceOver.  Adding the "OK" button does change what's announced, namely "1 other item" is mentioned.

chosww:
I think for now, having a button to close the modal is good enough, since we have nothing to use as a label in the modal. To improve user experience, I suggest setting focus on the ok button when the modal opens up.

chosww:
This works as expected for now, but setting focus on onClick event feels little weird to me, especially to the button that opens up the modal. For the other modal components, I used componentDidUpdate method to set the focus when show property changes from false to true (When the modal opens up). I suggest move the focus logic inside of the ConfirmDeleteAllModal component, and make setting focus responsibility of the modal.

the-t-in-rtf:
Good idea, and done.

the-t-in-rtf:
Done.  I don't like the styling of the OK button all that much, but I think we should hold off on doing anything further until we can properly refactor all the modals.

chosww:
Yeah, this is temporary design, and we actually have a new design for the share modal, so I think it's fine as it is now.

PR #276, opened 2021-09-29
C2LC-490: Announce scene name in scene announcements.

the-t-in-rtf:
See [C2LC-490](https://issues.fluidproject.org/browse/C2LC-490) for details.


chosww:
The announcement should include which world the character is in, I suggest making changes to the beginning of the scene announcement like:

"Scene, in deep ocean", "Scene, in jungle safari" .. etc

the-t-in-rtf:
I did a bunch of manual testing of this, as far as I can see all announcements are as discussed both when manually announcing the scene, and when using VoiceOver.

src/Scene.js
```diff
@@ -201,7 +204,9 @@ class Scene extends React.Component<SceneProps, {}> {
                     numColumns,
                     numRows,
                     direction,
-                    relativeDirection: this.getRelativeDirection(xPos, yPos)
+                    relativeDirection: this.getRelativeDirection(xPos, yPos),
+                    world: this.props.world,
```
sbates-idrc, line 208: Please localize the world, rather than using the internal name directly.

src/Scene.js
```diff
@@ -216,7 +221,9 @@ class Scene extends React.Component<SceneProps, {}> {
                         xPos: columnLabel,
                         yPos: rowLabel,
                         direction,
-                        item: itemOnGridCell
+                        item: itemOnGridCell,
+                        world: this.props.world,
```
sbates-idrc, line 225: Please localize the world, rather than using the internal name directly.

src/Scene.js
```diff
@@ -227,7 +234,9 @@ class Scene extends React.Component<SceneProps, {}> {
                     numRows: this.props.dimensions.getHeight(),
                     xPos: columnLabel,
                     yPos: rowLabel,
-                    direction
+                    direction,
+                    world: this.props.world,
```
sbates-idrc, line 238: Please localize the world, rather than using the internal name directly.

the-t-in-rtf:
Thanks for catching and fixing the raw world name issue.

PR #277, opened 2021-09-29
C2LC-487: Improve screen reader experiences of modal components

chosww:
This work includes changes we proposed on C2LC-488 as well. Also, there's no changes for the share modal in this PR.

See [C2LC-487](https://issues.fluidproject.org/browse/C2LC-487) and [C2LC-488](https://issues.fluidproject.org/browse/C2LC-488) for details.


the-t-in-rtf:
I tested this and compared directly with the 0.9.1 preview site using Safari and VoiceOver.  All dialogues (except the share, as noted) are announced as dialogues.  There is always an object focused, generally the button to save changes or close the dialogue.  I tested the world picker a few times to confirm that the initial focus is on the selected world, which I liked.

src/ConfirmDeleteAllModal.js
```diff
@@ -46,6 +48,16 @@ class ConfirmDeleteAllModal extends React.Component<ConfirmDeleteAllModalProps,
             </Modal>
         );
     }
+
+    componentDidUpdate (prevProps: ConfirmDeleteAllModalProps) {
+        if (prevProps.show !== this.props.show && this.props.show) {
```
the-t-in-rtf, line 53: We use this pattern often, and we have a general version of this in App.js.  Longer term I think we should move that function to Utils.js and replace all usages.  I will create a new ticket for that cleanup, but it would be nice to leave a TODO here.

src/ConfirmDeleteAllModal.js
```diff
@@ -46,6 +48,16 @@ class ConfirmDeleteAllModal extends React.Component<ConfirmDeleteAllModalProps,
...
+            const deleteAllButton = document.getElementById('deleteAll-button');
+            if (deleteAllButton) {
+                deleteAllButton.focus();
+            }
+
```
the-t-in-rtf, line 58: This line should probably be removed.

src/KeyboardInputModal.js
```diff
@@ -232,13 +235,20 @@ class KeyboardInputModal extends React.Component<KeyboardInputModalProps, Keyboa
...
                 keyBindingsEnabled: this.props.keyBindingsEnabled,
                 keyboardInputSchemeName: this.props.keyboardInputSchemeName
             });
         }
+        if (prevProps.show !== this.props.show && this.props.show) {
+            const keyboardShortcutsToggle = document.getElementById('keyboardInputModal__toggle');
+            if (keyboardShortcutsToggle) {
+                keyboardShortcutsToggle.focus();
+            }
+
```
the-t-in-rtf, line 250: This kind of demonstrates my point about making a common utility, but again, we should do it in another ticket.  Please add a TODO here.  Also, the extra blank line is here too.

src/WorldSelector.js
```diff
@@ -117,6 +117,10 @@ class WorldSelector extends React.Component<WorldSelectorProps, WorldSelectorSta
     componentDidUpdate(prevProps: WorldSelectorProps, prevState: WorldSelectorState) {
         // When the modal first open up, remember the world at that time
         if (prevProps.show !== this.props.show && this.props.show) {
+            const selectedWorld = document.getElementById(`WorldSelector__input-world-${this.props.currentWorld}`);
```
the-t-in-rtf, line 120: Please leave another TODO here about replacing this with a common function.

the-t-in-rtf:
I created [https://issues.fluidproject.org/browse/C2LC-493](https://issues.fluidproject.org/browse/C2LC-493) to cover consolidating various similar focus management functions, I'd appreciate it if you could mention that in TODOs in the focus management functions you added here.  Once that's done I think this is ready for further review.

PR #280, opened 2021-10-01
C2LC-492: Clean up Keyboard Shortcut modal

chosww:
See [C2LC-492](https://issues.fluidproject.org/browse/C2LC-492) for details.


the-t-in-rtf:
The behaviour is improved when reading the current state, it now says "Keyboard shortcuts toggle, on, switch." instead of "Turn keyboard shortcuts on or off, off, switch".

The behaviour when changing state still isn't great, at least with Voiceover.  If I change from Off to On, it says, "On, keyboard shortcuts toggle, switch".  The same message we use when reading the first state would also make sense here if that's possible.

chosww:
@the-t-in-rtf, yeah that's how VoiceOver announces on toggle state change -- says changed state first like on or off, and then says the name of the switch. The behaviour is the same for other toggle switches, so I think that's something we have no control.

PR #281, opened 2021-10-04
C2LC-494: Always allow unhiding of hidden actions.

the-t-in-rtf:
See [C2LC-494](https://issues.fluidproject.org/browse/C2LC-494) for more details.


chosww:
The changes look good and works as expected.

PR #287, opened 2021-10-14
C2LC-493: Consolidated focus functions.

the-t-in-rtf:
See [C2LC-493](https://issues.fluidproject.org/browse/C2LC-493) for details.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -86,13 +86,6 @@ function extend(...toMerge:Object) {
     return merged;
 }
 
-function focusById (elementId: string) {
-    const element = document.getElementById(elementId);
-    if (element) {
-        element.focus();
-    }
-}
-
 function focusByQuerySelector (querySelector: string) {
     const element = document.querySelector(querySelector);
     if(element) {
@@ -100,13 +93,4 @@ function focusByQuerySelector (querySelector: string) {
     }
 }
 
-// Convenience function to focus on the first element with a given class, used
-// for keyboard shortcuts.
-function focusOnFirstElementWithClass (className: string) {
-    const elements = document.getElementsByClassName(className);
-    if (elements.length) {
-        elements[0].focus();
-    }
-}
-
-export { extend, focusById, focusByQuerySelector, focusOnFirstElementWithClass, generateId, makeDelayedPromise, generateEncodedProgramURL, getThemeFromString, getWorldFromString };
+export { extend, focusByQuerySelector, generateId, makeDelayedPromise, generateEncodedProgramURL, getThemeFromString, getWorldFromString };
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -86,5 +86,27 @@ function extend(...toMerge:Object) {
     return merged;
 }
 
+function focusById (elementId: string) {
+    const element = document.getElementById(elementId);
+    if (element) {
+        element.focus();
+    }
+}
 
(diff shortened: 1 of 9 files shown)


the-t-in-rtf:
I moved all variations on "find a thing and focus if it exists" to common functions in Utils.js.  The focus trap manager and program block editor needed more complex logic, so I left those alone.

the-t-in-rtf:
> I am wondering if we can just have one focus function that uses querySelector, instead of having 3 different variations? 

I see what you mean, [`querySelector` already focuses on the first element](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector), which I hadn't realised.

the-t-in-rtf:
> Also, I think we can add test case to cover the focus logic, just render a component with class/id in document in the test case, and call the function to check if focus is set on the component.

Done.  I ended up having to use a similar strategy as your modal work to get the test fixture to attach, glad we had that discussion.

src/Utils.test.js
```diff
@@ -37,3 +42,30 @@ test('Test extend', () => {
...
+        </div>,
+        // $FlowFixMe:  Flow is worried that document.body might not exist.
+        {attachTo: document.body}
+    );
+
+    const secondElement = testFixture.find(".second");
+
+    // We should be focused on the body by default.
+    // $FlowFixMe: Flow is worried that document.activeElement might not exist.
+    expect(document.activeElement.localName).toEqual('body');
```
chosww, line 62: Do we need to test this? As you commented, I think focus on body is default expected behaviour.

src/Utils.test.js
```diff
@@ -37,3 +42,30 @@ test('Test extend', () => {
     expect(extend({ a: 0 }, {b: 1}, { c: 2})).toEqual({ a: 0, b: 1, c: 2});
     expect(extend({ foo: { bar: { baz: true }}}, { foo: { new: true, bar: { new: true, baz: false}}})).toEqual({ foo: { new: true, bar: { new: true, baz: false}}});
 });
+
+test('Test focusByQuerySelector', () => {
+    // Test fixture with three focusable elements.
```
chosww, line 47: Let's add expect.assertions to make sure this test is checking expects.

src/Utils.test.js
the-t-in-rtf: I'm just making sure it's not focused on the element before I change focus.

src/Utils.test.js
the-t-in-rtf: Done.

src/Utils.test.js
the-t-in-rtf: It's a little superstitious, I know.

src/Utils.js
```diff
@@ -86,5 +86,11 @@ function extend(...toMerge:Object) {
     return merged;
 }
 
+function focusByQuerySelector (querySelector: string) {
```
sbates-idrc, line 89: Thanks very much for adding this function. It's really a good one to have.

Minor comment: could you rename the parameter to `selectors` please to match the parameter name of https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector to make it more transparent that it's directly using `Document.querySelector()`.

sbates-idrc:
I left one small comment on changing the name of the `focusByQuerySelector` parameter, but I just went ahead and made the change myself, so as not to delay merging of the PR.

PR #290, opened 2021-10-26
C2LC-536: Align character icon and program sequence vertically.

the-t-in-rtf:
See [C2LC-536](https://issues.fluidproject.org/browse/C2LC-536) for details.


src/ProgramBlockEditor.scss
```diff
@@ -134,6 +134,7 @@ button.btn.ProgramBlockEditor__program-deleteAll-button--pressed.ProgramBlockEdi
     display: flex;
     justify-content: center;
     padding: 0 0.5rem;
+    padding-top: 2.9rem;
```
chosww, line 137: Program sequence has padding-bottom: 1rem set to it, and would be great if we can just set same amount of bottom padding to this character column. We can make character icon to start at the bottom of the column, and give 1rem margin at the bottom of it by setting align-items: self-end. Also, let's remove padding rules in character-column and use padding-bottom: 1rem.

```
.ProgramBlockEditor__character-column {
    align-items: center;
    background-color: #30444E;
    border-right: 1px solid #19282F;
    border-radius: 3px 0 0 3px;
    box-shadow: 2px 0px 1px #8F8F8F;
    display: flex;
    justify-content: center;
    align-items: self-end;
    padding-bottom: 1rem;
```

chosww, line 137: With this padding top, the icon may look aligned with program sequence, but I think we should apply same style that's applied in the program sequence. The program sequence uses padding-bottom: 1rem to position itself, so what we can do here is to make the icon start from the bottom of the character column container, and then use bottom-padding: 1rem as well. Let's update .Program__character-column style to be:

```
.ProgramBlockEditor__character-column {
    align-items: center;
    background-color: #30444E;
    border-right: 1px solid #19282F;
    border-radius: 3px 0 0 3px;
    box-shadow: 2px 0px 1px #8F8F8F;
    display: flex;
    flex-direction: column;
    justify-content: flex-end;
    padding-bottom: 1rem;
}
```

the-t-in-rtf: This did not work for me.  Were there other changes required?

Screenshot 2021-11-16 at 11 46 34

chosww:
Hey @the-t-in-rtf, as we discussed on our tech channel, the previous style I suggest is incorrect and is not scaleable. So what we should do is to apply same padding-top to the character Icon as the program sequence. Currently class ProgramBlockEditor__program-sequence-scroll-container is applying padding-top of 2.2rem, and class ProgramBlockEditor__program-sequence is applying another padding-top of 2.5rem. So we need to apply total 4.7rem of padding-top to the character Icon, and we need to change justify-content to center to make it always center the icon even when the container expands. The character column style should look like this:

```
.ProgramBlockEditor__character-column {
    align-items: center;
    background-color: #30444E;
    border-right: 1px solid #19282F;
    border-radius: 3px 0 0 3px;
    box-shadow: 2px 0px 1px #8F8F8F;
    display: flex;
    flex-direction: column;
    **justify-content: center;
    padding-top: 4.7rem;**
    padding-bottom: 1rem;
}
```

the-t-in-rtf:
@chosww I used your suggestion with a small tweak.  Using a `padding-top` of `4.7rem` ends up aligning with the bottom of the program sequence icons, since the padding is different than the character icon.  I subtracted `0.4rem` for the padding difference and it aligns with the middle in all but very high and low zoom levels.

chosww:
Thanks @the-t-in-rtf , padding-top of 4.7rem should be top align, but there's this light border around the icon that maybe hard to notice. With 4.3rem padding, it gives noticeable amount of gap in high zoom level as you said, but there's no gap when using 4.7rem padding. I think we should use 4.7rem padding.
Screen Shot 2021-12-10 at 10 03 39 AM

the-t-in-rtf:
@chosww,  I still see bottom alignment with 4.7rem at normal zoom levels even, maybe we can pair on it and discuss in a meeting.

sbates-idrc:
We have removed the character from the program editor. Closing.

PR #291, opened 2021-10-27
C2LC-499: Implement a modal component

chosww:
See [C2LC-499](https://issues.fluidproject.org/browse/C2LC-499) for details.


the-t-in-rtf:
I see that the radio button selection is fixed, but discovered a new problem.  In the current react bootstrap implementation, when the modal opens, voiceover announces that it's a dialog and summarises its contents.  With the new design, only the single radio button is announced with no further context.

In the old design, the title and description of the modal are also reachable and readable using VoiceOver's arrow key functionality.   This is not the case with the new design, so not only do you not hear the context when the modal opens, but you can't navigate back to it using VoiceOver and the arrow keys.

src/ModalWithFooter.js
```diff
@@ -0,0 +1,59 @@
...
+type ModalWithFooterProps = {
+    show: boolean,
+    focusElementSelector: string,
+    focusOnCloseSelector: string,
+    ariaLabel?: string,
+    ariaLabelledById?: string,
+    ariaDescribedById?: string,
+    children?: any,
+    onClose: () => void,
+    buttonProperties: Array<{label: string, onClick: () => void, isPrimary: boolean}>
```
the-t-in-rtf, line 18: Seems like we'd have to write less boilerplate if `isPrimary` could be optional, so that we'd only have to specify it for the primary button.

I do kind of wonder about this strategy WRT three button layouts, but as we only have 1 and 2 button modals at the moment, it's fine to have `isPrimary` for the primary and know that a secondary button omits that or has it set to false.

src/ConfirmDeleteAllModal.js
```diff
@@ -18,43 +18,31 @@ type ConfirmDeleteAllModalProps = {
...
-                    </div>
-                </Modal.Body>
-            </Modal>
+                focusElementSelector={'#DeleteAll-done'}
+                focusOnCloseSelector={'.ProgramBlockEditor__program-deleteAll-button'}
+                ariaLabel={this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.confirmButton'})}
+                onClose={this.props.onCancel}
+                buttonProperties={
+                    [
+                        {label: this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.cancelButton'}), onClick: this.props.onCancel, isPrimary: false},
```
the-t-in-rtf, line 29: I still think we could find a way to not have to indicate that `isPrimary` is `false`.

the-t-in-rtf:
I left a few comments and questions, but on the whole this seems ready for further review.

src/ConfirmDeleteAllModal.js
chosww: I think it is safe to assign false this way, because we are checking isPrimary property to decide className for a specific button, and if we make isPrimary optional, the logic for the decision would still work, but I think using negation on undefined little unsafe.

src/Modal.js
```diff
@@ -0,0 +1,170 @@
+// @flow
+
+import React from 'react';
+import classNames from 'classnames';
+import './Modal.scss';
+
+type ModalProps = {
+    show: boolean,
+    focusElementSelector: string,
```
sbates-idrc, line 9: I think rename `focusElementSelector` to `focusOnOpenSelector` to match `focusOnCloseSelector` and to make the role clearer.

src/Modal.js
```diff
@@ -0,0 +1,170 @@
...
+    ariaDescribedById?: string,
+    children?: any,
+    onClose: () => void
+};
+
+class Modal extends React.Component<ModalProps, {}> {
+    modalRef: { current: null | Element };
+    lastFocus: any;
+    ignoreFocusChanges: boolean;
+    constructor(props: ModalProps) {
```
sbates-idrc, line 22: Put an empty line before the `constructor`

src/Modal.js
```diff
@@ -0,0 +1,170 @@
...
+        }
+        this.ignoreFocusChanges = false;
+        return (document.activeElement === element);
+    };
+
+    handleFocusTrap = (event: Event) => {
+        if (this.ignoreFocusChanges) {
+            return;
+        }
+        // Add guard cases to check if properties exist
```
sbates-idrc, line 70: What does this comment mean?

src/Modal.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    handleOnClose = () => {
+        // $FlowFixMe: flow thinks document.body can be null
+        document.body.classList.remove('modal-opened');
+        this.props.onClose();
+    }
+
+    // rename it to be handleKeyDown
+    handleOnPressEscapeKey = (event: Event) => {
```
sbates-idrc, line 92: Rename `handleOnPressEscapeKey` to `handleKeyDown`

src/Modal.js
```diff
@@ -0,0 +1,170 @@
+// @flow
+
+import React from 'react';
+import classNames from 'classnames';
+import './Modal.scss';
+
+type ModalProps = {
```
sbates-idrc, line 7: Our Modal component is based on the W3C example code and we should add a note to acknowledge that.

* https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/js/dialog.js
* https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document

src/Modal.js
chosww: I think I left a note for myself and this comment seems meaningless now. I removed the comment.

src/Modal.js
chosww: Thank you for the suggestion, I did not think about that. Please let me know if the acknowledgement I wrote is not enough.

sbates-idrc: The license includes a template to use: "This software or document includes material copied from or derived from [title and URI of the W3C document]. Copyright © [YEAR] W3C® (MIT, ERCIM, Keio, Beihang)." Please use that:

```
// This software or document includes material copied from or derived from
// WAI-ARIA Authoring Practices 1.1 Modal Dialog Example
// https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html
// Copyright © 2019 W3C® (MIT, ERCIM, Keio, Beihang).
// Source code: https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/js/dialog.js
// License: https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
```

sbates-idrc: And please also add a note and link to the license to the table at the bottom of our README.

README.md
```diff
@@ -72,3 +72,4 @@ use in Weavly:
 | [github.com/sass/node-sass](https://github.com/sass/node-sass) | MIT License |
 | [github.com/Tonejs/Tone.js](https://github.com/Tonejs/Tone.js) | [MIT License](https://github.com/Tonejs/Tone.js/blob/dev/LICENSE.md)|
 | [loading.io](https://loading.io/) | [Loading.io Free License](https://loading.io/license/#free-license) |
+| [w3c.org](https://www.w3.org/) | [W3C License](https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document) |
```
sbates-idrc, line 75: For the link in the 1st column, please link to the doc:

```
[WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/)
```

src/ModalWithFooter.scss
```diff
@@ -0,0 +1,41 @@
...
+
+    &:focus {
+        box-shadow:
+            0 0 0 0.1rem #D5D8DD,
+            0 0 0 $c2lc-focus-indicator-width black;
+    }
+}
+
+button.btn.ModalWithFooter__secondaryButton {
+    margin-left: 5rem;
```
sbates-idrc, line 33: What is the purpose of this `margin-left` on the secondary button?

chosww: The new modal design has some spacing left to the cancel buttons, but I see this can be an issue when there's only one Primary button, or multiple secondary buttons. I think putting padding on the footer container makes more sense.

src/ConfirmDeleteAllModal.js
```diff
@@ -18,43 +18,31 @@ type ConfirmDeleteAllModalProps = {
...
+                ariaLabel={this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.confirmButton'})}
+                onClose={this.props.onCancel}
+                buttonProperties={
+                    [
+                        {label: this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.cancelButton'}), onClick: this.props.onCancel},
+                        {id: 'DeleteAll-done', label: this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.confirmButton'}), onClick: this.props.onConfirm, isPrimary: true}
+                    ]
+                }>
+                <ModalHeader
+                    id='WorldSelector'
```
sbates-idrc, line 34: Wrong id. This is the Delete All Modal.

src/ProgramBlockEditor.test.js
```diff
@@ -109,7 +109,7 @@ function createMountProgramBlockEditor(props) {
 }
 
 function confirmDeleteAllModalIsOpen(programBlockEditorWrapper): boolean {
-    return programBlockEditorWrapper.exists('.ConfirmDeleteAllModal');
+    return programBlockEditorWrapper.exists('.Modal__container.active');
```
sbates-idrc, line 112: This checks that a modal has been opened. Not that the Delete All modal specifically has been opened.

src/Themes.scss
```diff
@@ -673,8 +673,8 @@ body.gray-theme {
         }
     }
 
-    button.btn.ModalFooter__cancelButton,
-    button.btn.ModalFooter__doneButton {
+    button.btn.ModalWithFooter__secondaryButton,
+    button.btn.ModalWithFooter__primaryButton {
```
sbates-idrc, line 677: Until we remove the usage of ModalFooter (still used in the keybaord dialog), we'll need both sets of styles.

src/Themes.scss
```diff
@@ -1415,8 +1415,8 @@ body.contrast-theme {
...
```
sbates-idrc, line 1419: Until we remove the usage of ModalFooter (still used in the keybaord dialog), we'll need both sets of styles.

src/ConfirmDeleteAllModal.js
```diff
@@ -18,43 +18,31 @@ type ConfirmDeleteAllModalProps = {
...
-                </Modal.Body>
-            </Modal>
+                focusOnOpenSelector={'#DeleteAll-done'}
+                focusOnCloseSelector={'.ProgramBlockEditor__program-deleteAll-button'}
+                ariaLabel={this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.confirmButton'})}
+                onClose={this.props.onCancel}
+                buttonProperties={
+                    [
+                        {label: this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.cancelButton'}), onClick: this.props.onCancel},
+                        {id: 'DeleteAll-done', label: this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.confirmButton'}), onClick: this.props.onConfirm, isPrimary: true}
```
sbates-idrc, line 30: Let's use the full component name as the id prefix, and "confirm" rather than "done" to match our other naming: `ConformDeleteAllModal-confirm`

sbates-idrc:
The confirmation button for the World Selector should be "Save", rather than "Done".

src/ConfirmDeleteAllModal.js
```diff
@@ -18,43 +18,31 @@ type ConfirmDeleteAllModalProps = {
...
-                            className='ConfirmDeleteAllModal__option-button confirm'
-                            onClick={this.props.onConfirm}>
-                            <FormattedMessage id='ConfirmDeleteAllModal.confirmButton' />
-                        </Button>
-                    </div>
-                </Modal.Body>
-            </Modal>
+                focusOnOpenSelector={'#ConformDeleteAllModal-confirm'}
+                focusOnCloseSelector={'.ProgramBlockEditor__program-deleteAll-button'}
+                ariaLabel={this.props.intl.formatMessage({id: 'ConfirmDeleteAllModal.confirmButton'})}
```
sbates-idrc, line 25: I think this should be the title of the dialog "Delete Program". Either by using `ariaLabelledById`, as we do for the `WorldSeclector`, or using `this.props.intl.formatMessage({ id: 'ConfirmDeleteAllModal.title' })` directly.

sbates-idrc:
On iPad, I'm seeing the World Selector and the Delete Program dialog modals sitting too low on the page. They are centred on desktop, but on iPad, they are lower than centre vertically. The gap below the dialogs is smaller than the one above.

PR #294, opened 2021-11-02
C2LC-540: Migrated refresh, stop, and delete all buttons to use IconButton.

the-t-in-rtf:
See [C2LC-540](https://issues.fluidproject.org/browse/C2LC-540) for details.


src/App.js
```diff
@@ -1127,9 +1119,10 @@ export class App extends React.Component<AppProps, AppState> {
                                 </a>
                             </h1>
                             <IconButton
+                                className="App__header-keyboardMenuIcon"
                                 ariaLabel={this.props.intl.formatMessage({ id: 'KeyboardInputModal.ShowHide.AriaLabel' })}
                                 onClick={this.handleKeyboardModalToggle}
-                                onKeyDown={this.handleKeyboardMenuIconKeydown}
+                                onKeyDown={this.handleKeyboardModalToggle}
```
chosww, line 1125: For handleKeybaordModalToggle event handler, we can do the same as handleKeyDownWorldIcon, where it just sets showKeyboardModal state to true.

src/IconButton.js
```diff
@@ -28,8 +42,8 @@ class IconButton extends React.Component<IconButtonProps, {}> {
                 className={classes}
```
chosww, line 42: Screen Readers don't know that the IconButton is disabled and we need to inform that the button is disabled by using aria-disabled property. 
```
aria-disabled={this.props.disabled}
```

src/IconButton.test.js
```diff
@@ -0,0 +1,103 @@
...
+
+            expect(clickHandler.mock.calls.length).toBe(0);
+
+            wrapper.simulate('click');
+
+            expect(clickHandler.mock.calls.length).toBe(1);
+        });
+    });
+
+    describe('Given disabled is true', () => {
```
chosww, line 67: Let's add test cases to check when disabled is true, wrapper has aria-disabled property set to true as well.

src/ProgramBlockEditor.scss
```diff
@@ -73,56 +73,54 @@
     }
 }
 
-button.btn.ProgramBlockEditor__program-deleteAll-button {
+.ProgramBlockEditor__program-deleteAll-button {
```
chosww, line 76: We can remove all the styles, except margin-left from the deleteAll-button, as its style is from IconButton.

chosww: Including its focus style.

src/ProgramBlockEditor.scss
```diff
@@ -73,56 +73,54 @@
...
     margin-left: 1.5rem;
-    padding: 0.5rem;
 
     &:focus {
         box-shadow: 0 0 0 $c2lc-focus-indicator-width black;
     }
 }
 
-button.btn.ProgramBlockEditor__program-deleteAll-button--pressed {
+.ProgramBlockEditor__program-deleteAll-button--pressed {
```
chosww, line 88: We can remove this style, as this className doesn't exist in ProgramBlockEditor.

src/ProgramBlockEditor.scss
```diff
@@ -73,56 +73,54 @@
...
     background-color: #000;
 }
 
-button.btn.ProgramBlockEditor__program-deleteAll-button.ProgramBlockEditor__program-deleteAll-button--disabled,
-button.btn.ProgramBlockEditor__program-deleteAll-button.ProgramBlockEditor__program-deleteAll-button--disabled:hover,
-button.btn.ProgramBlockEditor__program-deleteAll-button.ProgramBlockEditor__program-deleteAll-button--disabled:active {
+.ProgramBlockEditor__program-deleteAll-button.ProgramBlockEditor__program-deleteAll-button--disabled,
```
chosww, line 92: We can remove all specific state styles to the deleteAllButton, as we want it to be consistent with other IconButtons.

chosww: Keep .ProgramBlockEditor__program-deleteAll-button-svg style, and let'e remove all other styles.

chosww:
I see that you kept button specific css rules after you have changed them to use IconButton component. With changing control buttons to use IconButton, we don't need to apply any specific button styles to each of them, but we still need to keep styles applied to SVGs, as Icons have different Fill and Stroke properties. I see some issues for deleteAll button icon in the grayscale and high contrast theme, and refresh button in high contrast theme (cause by border specified on the button). I Also see that focus style on the stop button is still not centred.

src/IconButton.test.js
the-t-in-rtf: Done.

src/IconButton.js
the-t-in-rtf: Done.

src/App.js
the-t-in-rtf: I see what you mean, the modal itself handles setting it the other way.  Done.

src/ProgramBlockEditor.scss
the-t-in-rtf: This breaks the contrast styles, but I'll move the relevant stuff there.

the-t-in-rtf: Huh, guess the contrast styles were broken before.  I'll fix those.

src/ProgramBlockEditor.scss
the-t-in-rtf: I went through and got rid of all the ones that either singly or in combination were for the unused `pressed` variant.

the-t-in-rtf:
There were a few cases in which styles were necessary because the background for a given button was different than the rest of the icon buttons, but I removed as many redundant styles as I could in all themes and fixed the issues you reported.  This should be ready for further review.

chosww:
Thanks Tony, all the changes look great, except styles on StopButton. I see an issue in default style for the stop button in the default, light and dark theme, where it looks like has an extra border from its background colour. I also think that setting background colour for the stop button and using padding-top to make the icon looks centered is not a good approach. I have an alternative idea, which is to update Stop and Play button SVGs to just contain the square and triangle, and we will style the outer ring with css. We can safely use box-shadow to draw the outer circle, but we need nested box-shadow for the focus styles, as focus style applies a box-shadow on the buttons. Something like this:

```
button.btn.PlayButton:focus, .StopButton:focus {
    border: none;
    box-shadow:
        inset 0 0 0 0.3rem #67717E,
        0 0 0 $c2lc-focus-indicator-width white;
}
```

the-t-in-rtf:
I updated the stop icon as suggested to remove the outer circle.  I disentangled the styles from the PlayButton styles and made sure there were specific enough rules to override the IconButton styles that differed.  I tested it in all themes and tried to make it match the play button for both disabled and enabled styles.

The one thing I wanted to mention is that I did still have to adjust the font-size to get the icon to centre itself vertically.  I scaled the font-size in proportion to the relative sizes of the play and stop buttons and it seems well-aligned.

@chosww, this should be ready for further prereview.

src/StopButton.scss
```diff
@@ -1,21 +1,41 @@
...
 
-.StopButton-svg {
-    width: 2.6rem;
-    height: 2.6rem;
+.StopButton {
+    height: 2rem;
+    width: 2rem;
+    font-size: 1.2rem;
+
+    box-shadow: 0 0 0 0.2rem black;
```
chosww, line 8: We need to update any usage of "black" in StopButton to be "#1E1E1E" to match designs in Figma.

chosww:
Thanks @the-t-in-rtf, styling the outer circle of the stop button using css looks good. It just need some minor style updates, and as I talked about in our chat, I think it makes sense to wait C2LC-485 to be merged into dev branch first, and update your work with dev branch for the correct hover style for stop and play buttons. I also see an issue with hover style of deleteAll button in high contrast theme.

the-t-in-rtf:
I fixed the colours for the stop button.  I also fixed the hover styles for the deleteAll button, which needed a more specific CSS selector to beat the underlying IconButton styles.

src/Themes.scss
```diff
@@ -286,14 +286,29 @@ body.light-theme {
...
+
+        svg rect {
+            fill: #3A3F46;
+        }
+
+        &:focus {
+            box-shadow:
+                0 0 0 0.2rem #3A3F46,
+                0 0 0 0.4rem #D5D8DD,
+                0 0 0 0.6rem black;
```
chosww, line 308: We should use #1E1E1E for Stop button in Themes.scss too.

chosww:
DeleteAll Button's hover style in high contrast theme got fixed, but I found another issue with refresh button's hover style in high contrast theme. Also, we use colour black for fill and stroke of many different icons, but all of them should be #1E1E1E --  I went through Figma, and we never use colour black for anywhere in the coding environment, so would be good to make a scss variable with name c2lc-black and update all usage of black to use the variable, but that should be in another ticket.

src/Themes.scss
the-t-in-rtf: Done.

the-t-in-rtf:
I fixed the hover styles for the scene refresh in high contrast.  Similar to the world picker, on hover the colour of the icon changes to the colour of the background.  I added a thinner outline when the scene refresh is hovered but not focused.


I fixed the stray 'black' reference and created [C2LC-544](https://issues.fluidproject.org/browse/C2LC-544) to cover the rest.

the-t-in-rtf:
@chosww, this should now include the fixes you made for C2LC-485, and should be ready for review.

As discussed in the chats, the styles for the pause button also needed work.  The pause styles can't exactly match the play/stop buttons, as the "pause" bars are cut out of the inner circle, revealing the colour used for the outer ring.  I did the best I could with those for now, IMO we should convert that to an IconButton in the next sprint.

sbates-idrc:
I've merged the latest `develop-1.0` into `C2LC-540` and made changes to remove the `onKeyDown` callback property from `IconButton`. Please merge my `C2LC-540` branch into yours to update this PR with those changes: https://github.com/sbates-idrc/c2lc-coding-environment/commits/C2LC-540

sbates-idrc:
I'm seeing some small outstanding styling issues:

* The Delete All button has hover styling (orange) in the Default, Light, and Dark themes when disabled, when it should not (see the Refresh button)
* We have lost some spacing between the Stop button and the speed slider
* The focus indicator on the disabled Stop button is too small in the Dark theme

Disabled Stop button focus in Default theme:

image

Disabled Stop button focus in Dark theme:

image

sbates-idrc:
This branch includes the change from C2LC-485 to make the background of the Play area lighter. But it doesn't include other changes that need to be made along with that colour change. In particular, with only the change to the background, the Share button now has very low contrast with the background. Please switch the Play area background back to what it is in `develop-1.0`.

Develop-1.0:

image

C2LC-540:

image

the-t-in-rtf:
@sbates-idrc, I believe I have incorporated all the suggested changes.

PR #295, opened 2021-11-02
C2LC-462: Removed screenshot from README.

the-t-in-rtf:
See [C2LC-462](https://issues.fluidproject.org/browse/C2LC-462) for details.


chosww:
The change looks good.

PR #296, opened 2021-11-08
C2LC-539: Platform detection and platform-specific wording for key bindings.

the-t-in-rtf:
See [C2LC-539](https://issues.fluidproject.org/browse/C2LC-539) for details.


src/messages.json
```diff
@@ -154,6 +155,7 @@
         "KeyboardInputModal.KeyIcons.Alt": "Alt",
         "KeyboardInputModal.KeyIcons.Control": "Ctrl",
         "KeyboardInputModal.KeyIcons.Shift": "Shift",
+        "KeyboardInputModal.KeyIcons.Option": "⌥",
```
chosww, line 158: I started to wonder do we really need to use intl to get the KeyIcons as they all seem to be international?

chosww:
Other than the question I have, the changes look good and works fine on MacBook, iPhone, iPad, Windows and Chromebook.

src/messages.json
the-t-in-rtf: It's potentially useful for countries that tend to use local abbreviations for the common modifier keys.  I couldn't find any French Canadian keyboards that went so far as to translate "control", et cetera, but I did find German layouts that were translated.   However, even in that case, it varies by keyboard, the Apple keyboard for the German market still uses the English terms.  You'd really have to assume that modern computer users would be used to the English terms for their platform, so probably we could just use that. For the letter keys, internationalisation seems less useful and potentially dangerous, as you never want to get in a situation where the key code doesn't match the label.

I could see moving these to a constant array in `KeyboardInputSchemes`, and it wouldn't take long.  I'll wait to get Simon's input first.

PR #297, opened 2021-11-11
C2LC-531:  Convert keyboard input modal to use ModalWithFooter.

the-t-in-rtf:
See [C2LC-531](https://issues.fluidproject.org/browse/C2LC-531) for details.

(This should be reviewed after C2LC-499 is merged.)


chosww:
Hey @the-t-in-rtf, so there's just a few css rules we need to change to complete this work, and styles we want to change are all in KeyboardInputModal.scss. 

- We can remove .KeyboardInputModal rule
- We can remove display: flex and any flex related rules from .KeyboardInputModal__content, and need to set width and max-height with overflow: auto rules; we want scrollbar in case content of the modal gets too length, but scroll bar to only work in the content container
- Move margin-top set on .KeyboardInputModal__content to .KeyboardInputModal__content__toggleBar to make scrolling experience more natural

Here's styles for .KeyboardInputModal__content I tried and worked well:

```
.KeyboardInputModal__content {
    overflow: auto;
    width: 45rem;
    max-height: 30rem;
    padding: 0;
}
```

the-t-in-rtf:
Thanks, @chosww.  The new styles definitely fix the side-to-side jumping.  However, the old modal was top-aligned as well.  This meant that although the save and cancel buttons jumped, the toggle didn't.   This seems like better behaviour, as the control you just activated shouldn't move away from where you're looking and where your fingers and/or mouse are.

I can think of a few strategies to address this:

1. Top-align all modals.
2. Top-align just this modal  (would require updating the modal to support a passed `className`).
3. Fix the height (this results in a very large empty space when shortcuts are disabled).
4. Always display the shortcuts, but with a greyed out style when keyboard shortcuts are disabled.

I kind of like the last one, as it completely reduces the vertical "jump" for all buttons, and also provides useful feedback on what keys would be supported if you were to turn them on.

the-t-in-rtf:
Now that C2LC-499 has been merged, this should now be much cleaner.  We confirmed that Simon's suggestion was good for Sepi, so I made that change to fix the vertical jumping when toggling keyboard shortcuts.  As discussed I also stopped disabling the dropdown when keyboard shortcuts are disabled.  This should be ready for review.

src/KeyboardInputModal.scss
```diff
@@ -1,9 +1,7 @@
 @import '_variables';
 
 .KeyboardInputModal__content {
-    overflow: auto;
```
chosww: Modal content is where we should place the scrollbar I think. Only containers in the modal those should always visible without scrolling should be the header and footer.

src/KeyboardInputModal.scss
```diff
@@ -1,9 +1,7 @@
...
     width: 45rem;
-    max-height: 30rem;
```
chosww: I suggest setting max-height back on the content container, and set it to use dynamic height I suggested on C2LC-531.

```
max-height: calc(100vh - 15rem);
```

src/KeyboardInputModal.scss
chosww: I had a chat with Simon, and they think it might be worth to render the control outside of the scrollable container. Let's talk about potential options during next meeting.

src/KeyboardInputModal.scss
the-t-in-rtf: I used the suggested approach.

chosww:
Recent dev-1.0 branch includes ModalBody component, which have rules for max-width and max-height for modal content. Please wrap content of KeyboardShortcutModal with ModalBody component. You can see an example on either WorldSelector or Delete Program Modal.

the-t-in-rtf:
Done, I'll update the others as well.

sbates-idrc:
I've made the changes that we discussed yesterday on this branch:

https://github.com/sbates-idrc/c2lc-coding-environment/commits/C2LC-531

Please take a look and let me know if it looks good. Thanks.

the-t-in-rtf:
I merged, looked at the changes, and tested.  I've pushed an update with your changes, thanks @sbates-idrc .

PR #298, opened 2021-11-11
C2LC-530: Convert actions simplification menu to modal.

the-t-in-rtf:
See [C2LC-530](https://issues.fluidproject.org/browse/C2LC-530) for details.


the-t-in-rtf:
This is ready for preliminary review, @chosww.  I'm not perfectly happy with the styles of the icons, I could use your help tidying that up.

the-t-in-rtf:
I also am not terribly happy with the appearance at 200% zoom level.  I tried to get scrolling working reasonably, but couldn't find a way to keep the header/footer onscreen at high zoom without making it too small at normal size.

chosww:
> I also am not terribly happy with the appearance at 200% zoom level. I tried to get scrolling working reasonably, but couldn't find a way to keep the header/footer onscreen at high zoom without making it too small at normal size.

We could use more dynamic height value for the modal contents. I tried using calc(100vh - 15rem) as the height for .ActionsSimplificationModal__menu, it could support various zoom levels well.

chosww:
I noticed that in Figma design, we are using term "step" instead of "square". I think word step makes more sense, but would be good to double check with Sepideh. We are using word square for the announcement too, so if step is more appropriate, we would need to change the announcement and aria-labels.

src/App.js
```diff
@@ -1202,13 +1234,21 @@ export class App extends React.Component<AppProps, AppState> {
...
-                            editingDisabled={this.editingIsDisabled()}
-                            programSequence={this.state.programSequence}
-                            intl={this.props.intl}
-                        />
+                        <div className='App__ActionsMenu__header'>
+                            <h2 className='App__ActionsMenu__header-heading'>
+                                <FormattedMessage id='ActionsMenu.title' />
+                            </h2>
+
+                            <div className="App__ActionsMenu__toggle-button"
```
chosww, line 1242: We need aria-label for this button.

src/ActionsSimplificationModal.js
```diff
@@ -0,0 +1,186 @@
...
+        const saveButtonProperties = {
+            id: 'ActionSimplificationModal-done',
+            isPrimary: true,
+            label: this.props.intl.formatMessage({ id: 'ActionsSimplificationModal.save'},),
+            onClick: this.saveChanges
+        };
+        return (
+            <ModalWithFooter
+                show={this.props.show}
+                focusOnOpenSelector={'#ActionSimplificationModal-done'}
```
chosww, line 108: I think setting focus on the first item in the list might be better experience. Since we have unique id for each item, we could use the id to set focus.

src/ActionsSimplificationModal.js
```diff
@@ -0,0 +1,186 @@
...
+    constructor(props: ActionsSimplificationModalProps) {
+        super(props);
+        this.state = {
+            allowedActions: this.props.allowedActions
+        }
+    }
+
+    render() {
+        const cancelButtonProperties = {
+            label: this.props.intl.formatMessage({ id: 'ActionsSimplificationModal.cancel'},),
```
chosww, line 95: We could clean up this syntax a little, by removing , inside of formatMessage method.

src/ActionsMenuItem.js
```diff
@@ -17,43 +22,60 @@ type ActionsMenuItemProps = {
...
 
-        // If we're used, show one message. If we're not, show another that differs based on `isAllowed`.
-        const showHideLabelKey = this.props.isUsed ? "ActionsMenu.item.usedItemToggleLabel" : "ActionsMenu.item.unusedItemToggleLabel";
-        const showHideLabel = this.props.intl.formatMessage(
-            { id: showHideLabelKey },
-            { action: actionName, commandName: commandName }
-        );
-
-        const showHideAriaLabelKey = this.props.isUsed ? "ActionsMenu.item.usedItemToggleAriaLabel" : "ActionsMenu.item.unusedItemToggleAriaLabel";
+        const showHideAriaLabelKey = this.props.isUsed ? "ActionsMenuItem.usedItemToggleAriaLabel" : "ActionsMenuItem.unusedItemToggleAriaLabel";
```
chosww, line 37: With the changes to make simplification options check/uncheckable when they are used, we don't need to use "usedItemToggleAriaLabel" I think, as used option can be hidden.

chosww:
Thanks @the-t-in-rtf for your work, I checked Figma design and we have a checkbox with label movements at the top of the simplification modal, and was wondering if we decided not to include that fo 530? If we decided to include that, we should include a div inside of . ActionsSimplificationModal__menu div, and wrap the command checkboxes in a separate div and apply same background colour as the modal title.

the-t-in-rtf:
> We could use more dynamic height value for the modal contents. I tried using calc(100vh - 15rem) as the height for .ActionsSimplificationModal__menu, it could support various zoom levels well.

Thanks, that works well from 75% to far beyond the zoom levels we usually test at.

src/ActionsSimplificationModal.js
the-t-in-rtf: I tried this but it intermittently focuses without indicating anything visually.

src/ActionsSimplificationModal.js
the-t-in-rtf: Good catch.

the-t-in-rtf:
> Thanks @the-t-in-rtf for your work, I checked Figma design and we have a checkbox with label movements at the top of the simplification modal, and was wondering if we decided not to include that for 530? If we decided to include that, we should include a div inside of . ActionsSimplificationModal__menu div, and wrap the command checkboxes in a separate div and apply same background colour as the modal title.

I think we discussed previously leaving it out until we have things other than commands, i.e. loops or functions.

src/ActionsMenuItem.js
the-t-in-rtf: Fixed.

the-t-in-rtf:
> I noticed that in Figma design, we are using term "step" instead of "square". I think word step makes more sense, but would be good to double check with Sepideh. We are using word square for the announcement too, so if step is more appropriate, we would need to change the announcement and aria-labels.

I fixed the wording for the movement actions.

chosww:
> > This is ready for preliminary review, @chosww. I'm not perfectly happy with the styles of the icons, I could use your help tidying that up.
> 
> Looking at Figma design, the icons seem to have box-shadow around them to make their style more sticking out. Current deployed version's icons look too flat to me as well, would styling them with box-shadow make you feel better about the style?

I figured out a good box-shadow rule for the icons, let's apply this style in .ActionsMenuItem__icon:

```
box-shadow: 0 3px 4px 0 #1e1e1e, -1px 2px 4px
```

src/ActionsSimplificationModal.js
chosww: There's another same syntax issue for the saveButtonProperties.

src/Themes.scss
```diff
@@ -86,6 +98,12 @@ body.light-theme {
         }
     }
 
+    .ActionsMenuItem__icon--backward1, .ActionsMenuItem__icon--backward2, .ActionsMenuItem__icon--backward3 {
+        &:not(.ActionsMenuItem__icon--disabled) {
```
chosww, line 102: We don't need the not selector anymore.

src/Themes.scss
```diff
@@ -108,6 +132,12 @@ body.light-theme {
         }
     }
 
+    .ActionsMenuItem__icon--left45, .ActionsMenuItem__icon--left90, .ActionsMenuItem__icon--left180 {
+        &:not(.ActionsMenuItem__icon--disabled) {
```
chosww, line 136: We don't need the not selector anymore.

src/Themes.scss
```diff
@@ -119,6 +149,12 @@ body.light-theme {
         }
     }
 
+    .ActionsMenuItem__icon--right45, .ActionsMenuItem__icon--right90, .ActionsMenuItem__icon--right180 {
+        &:not(.ActionsMenuItem__icon--disabled) {
```
chosww, line 153: We don't need the not selector anymore.

src/Themes.scss
```diff
@@ -1610,6 +1670,37 @@ body.contrast-theme {
...
+        }
+
+        &:hover:focus {
+            box-shadow:
+                0 0 0 0.1rem white,
+                0 0 0 0.2rem black;
+        }
+    }
+
+    .ActionsMenuItem__icon:not(.ActionsMenuItem__icon--disabled) {
```
chosww, line 1690: We don't need the not selector anymore.

chosww (commented):
Thanks @the-t-in-rtf for all the updates, I left a few more comments based on the conversation this morning with Simon and Sepideh. We should also figure out good structure for the input checkboxes tomorrow, so there maybe more change suggestions after the meeting tomorrow.

src/ActionsSimplificationModal.js
the-t-in-rtf: Fixed.

src/messages.json
```diff
@@ -9,21 +9,20 @@
         "ActionsMenu.toggleActionsMenu": "configure available actions",
         "ActionsMenuItem.action.hide": "Hide",
         "ActionsMenuItem.action.show": "Show",
-        "ActionsMenuItem.command.backward1": "Move backward 1 square",
```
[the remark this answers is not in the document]
the-t-in-rtf: Done.

src/ActionsMenuItem.js
```diff
@@ -15,45 +20,72 @@ type ActionsMenuItemProps = {
...
-        const showHideAriaLabelKey = this.props.isUsed ? "ActionsMenu.item.usedItemToggleAriaLabel" : "ActionsMenu.item.unusedItemToggleAriaLabel";
-        const showHideAriaLabel = this.props.intl.formatMessage(
-            { id: showHideAriaLabelKey },
-            { action: actionName, commandName: commandName }
+        const textClassNames = classNames(
+            'ActionsMenuItem__text',
+            !this.props.isAllowed && 'ActionsMenuItem__text--disabled'
         );
 
+        const iconClassNames = classNames(
```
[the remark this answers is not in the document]
the-t-in-rtf: We still need two classnames for the icon, so I just got rid of the disabled classname.

src/ActionsMenuItem.js
```diff
@@ -15,45 +20,72 @@ type ActionsMenuItemProps = {
...
+            'ActionsMenuItem__icon',
+            'ActionsMenuItem__icon--' + this.props.itemKey,
+            !this.props.isAllowed && 'ActionsMenuItem__icon--disabled'
+        );
+
+        // $FlowFixMe: Flow is confused about what itemKey is.
+        let icon = null;
```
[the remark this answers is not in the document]
the-t-in-rtf: Fixed.

src/ActionsMenuItem.scss
```diff
@@ -1,33 +1,85 @@
...
+        margin: 0.25rem;
+        width: 2rem;
+        height: 2rem;
+    }
 }
 
-.ActionsMenuItem__text--disabled {
-    text-decoration: line-through;
-    color: #cccccc;
+.ActionsMenuItem__icon--disabled {
```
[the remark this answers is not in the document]
the-t-in-rtf: Done.

src/Themes.scss
the-t-in-rtf: I got rid of all of the not selectors.

src/ActionsSimplificationModal.js
```diff
@@ -0,0 +1,186 @@
...
+        const saveButtonProperties = {
+            id: 'ActionSimplificationModal-done',
+            isPrimary: true,
+            label: this.props.intl.formatMessage({ id: 'ActionsSimplificationModal.save'} ),
+            onClick: this.saveChanges
+        };
+        return (
+            <ModalWithFooter
+                show={this.props.show}
+                focusOnOpenSelector={'ActionsMenuItem:first-of-type'}
```
chosww, line 108: This needs to be a class selector, let's add . in front of the selector.

src/ActionsSimplificationModal.js
```diff
@@ -0,0 +1,186 @@
...
+                focusOnCloseSelector={'.App__ActionsMenu__toggle-button'}
+                onClose={this.props.onCancel}
```
chosww, line 110: We need to update the modal's state on onClose call, and onCancel handler from prop doesn't have the logic to update the state. Let's update onClose to use this.handleClickCancel. Maybe we could rename the handler to be handleOnCancel?

src/ActionsMenuItem.scss
```diff
@@ -1,33 +1,63 @@
...
-}
+    padding: 0.5rem 0.5rem;
+    border-radius: 0.2rem;
+    display: grid;
+    grid-template-columns: min-content 1fr min-content;
+    column-gap: 0.5rem;
+    row-gap: 0;
+    align-items: center;
 
-.ActionsMenuItem__checkbox {
```
chosww, line 7: I realized that clicking right on the checkbox doesn't check/uncheck the checkbox, we need to apply pointer-events: none on .ActionsMenuItem__checkbox. 

```
.ActionsMenuItem__checkbox {
    pointer-events: none; 
}
```

the-t-in-rtf: Good catch, and fixed.

src/ActionsSimplificationModal.js
the-t-in-rtf: Right, clicking cancel would have reset the state, but not clicking elsewhere or hitting escape.  It's fixed now.

src/ActionsSimplificationModal.js
the-t-in-rtf: Fixed.

the-t-in-rtf:
I updated to use the new ModalBody component and removed the redundant styles from the content.

sbates-idrc:
Tony, I've made some changes on my C2LC-530 branch as discussed. If you could check them over and test them, that would be awesome. Thanks!

https://github.com/sbates-idrc/c2lc-coding-environment/commits/C2LC-530

PR #300, opened 2021-11-17
C2LC-520: Converted share modal to use ModalWithFooter.

the-t-in-rtf:
See [C2LC-520](https://issues.fluidproject.org/browse/C2LC-520) for details.


the-t-in-rtf:
The existing share button handled a lot of things that are now part of the modal, so that also needed to be refactored.  I went ahead and collapsed that button into App.js and added the share icon from the Figma mockups.

src/App.scss
```diff
@@ -239,6 +239,52 @@ body {
...
+    background-color: #33717F;
+    border: 0;
+    border-radius: 0.2rem;
+    color: #FFFFFF;
+    font-weight: 400;
+    padding: 0.5rem 2rem;
+    width: 10rem;
+
+    display:flex;
+    grid-template-columns: 30% 70%;
```
chosww, line 253: This grid-template is not applying, because display is set to flex, but I think we don't need this template at all, since flex layout is handling it nicely. Let's remove this and remove white space on line 251 as well.

src/App.scss
```diff
@@ -239,6 +239,52 @@ body {
...
+    align-items: center;
+
+    &:active {
+        background-color: #ED972C;
+    }
+
+    &:focus {
+        box-shadow:
+            0 0 0 0.1rem #D5D8DD,
+            0 0 0 $c2lc-focus-indicator-width black;
```
chosww, line 264: Probably make sense to use #1E1E1E here.

src/ShareModal.js
```diff
@@ -0,0 +1,91 @@
...
+            label: this.props.intl.formatMessage({id: 'ShareModal.cancel'}),
+            onClick: this.props.onCancel,
+            isPrimary: false
+        }];
+
+        return(
+            <ModalWithFooter
+                show={this.props.show}
+                focusOnOpenSelector={'#ShareModal-copy'}
+                focusOnCloseSelector={'.ProgramBlockEditor__program-deleteAll-button'}
```
chosww, line 55: Need to update focusOnCloseSelector.

src/ShareModal.js
```diff
@@ -0,0 +1,91 @@
...
+                ariaLabel={this.props.intl.formatMessage({ id: 'ConfirmDeleteAllModal.title' })}
```
chosww, line 56: Need to update ariaLabel.

src/ShareModal.js
```diff
@@ -0,0 +1,91 @@
...
+        const buttonProperties = [{
+            label: this.props.intl.formatMessage({id: 'ShareModal.cancel'}),
+            onClick: this.props.onCancel,
+            isPrimary: false
+        }];
+
+        return(
+            <ModalWithFooter
+                show={this.props.show}
+                focusOnOpenSelector={'#ShareModal-copy'}
```
chosww, line 54: This Id doesn't exist, but it's working as expected, because the share button is first focusable item in the modal.

src/ShareModal.js
```diff
@@ -0,0 +1,91 @@
...
+                >
+                    <ShareIcon aria-hidden="true"/>
+                </ModalHeader>
+
+                <div className='ShareModal__content'>
+                    <p><FormattedMessage id='ShareModal.description1' /></p>
+                    <p><FormattedMessage id='ShareModal.description2' /></p>
+
+                    <div className='ShareModal__form'>
+                        <div className='ShareModal__form__URL__container'>
```
chosww, line 74: Was there a reason why you used a div to show the URL instead of text field? I think it would make sense to be able to manually highlight URL in the container if users want to. Also, by using text field, we don't need to apply custom styles for the container. Here's a suggestion:

```
<input className='ShareModal__form__URL' type="text" readOnly={true} value={document.location.href} />
```

chosww: If we decide to just use the text field I suggested, we can remove the ShareModal__form__URL__container and its child div, and just render the input element sibling of the share button.

src/ShareModal.scss
```diff
@@ -0,0 +1,56 @@
+@import '_variables';
+
+.ShareModal__content {
+    padding: 2rem;
+    border-bottom: 1px solid #E4E7EB;
+}
+
+.ShareModal__form {
+    display: flex;
```
chosww, line 9: Need to update display flex to grid to apply grid template style.

src/ShareModal.scss
```diff
@@ -0,0 +1,56 @@
...
+    grid-template-columns: 1fr min-content;    
+    column-gap: 1rem;
+}
+
+.ShareModal__form__URL__container {
+    border: 1px solid #9DA4AF;
+    border-radius: 0.2rem;
+    max-width: 50vh;
```
chosww, line 17: VH unit on width works, but doesn't make sense. Also, we can remove this container style if we just use a text field.

src/ShareModal.scss
```diff
@@ -0,0 +1,56 @@
...
+    padding: 0.5rem;
+}
+
+.ShareModal__form__URL {
```
chosww, line 21: From the Figma design, I think we want to use text-overflow: ellipsis to the visible URL part on the modal.

src/ShareModal.scss
```diff
@@ -0,0 +1,56 @@
...
+
+    background-color: #33717F;
+    border: 0;
+    border-radius: 0.2rem;
+    color: #FFFFFF;
+    font-weight: 400;
+    padding: 0.5rem 2rem;
+    width: 10rem;
+
+    display:flex;
```
chosww, line 37: The share button doesn't look centered, and I don't think we need some rules specified for this button. From Figma design, I see that the share button is the same width and height as footer buttons, and I think we can talk to Simon about what we talked, about making the footer button class more general and apply the style to this share button. If Simon thinks that's too out of scope, here's some style I tried and works well:

```
.ShareModal__form__copyButton {
    background-color: #33717F;
    border: 0;
    border-radius: 0.2rem;
    color: #FFFFFF;
    font-weight: 400;
    padding: 0.5rem 2rem;
    align-items: center;

    &:active {
        background-color: #ED972C;
    }

    &:focus {
        box-shadow:
            0 0 0 0.1rem #D5D8DD,
            0 0 0 $c2lc-focus-indicator-width black;
    }

    &:hover {
        background-color: #F5C58A;
        color: #1E1E1E;
    }
}
```

src/ShareModal.js
the-t-in-rtf: I was concerned that a text field would give the impression that the value was editable, or that people would mistakenly copy a partial URL manually.  We should discuss in the meeting tomorrow.

src/App.scss
the-t-in-rtf: Done.

src/App.scss
the-t-in-rtf: Done.

src/ShareModal.js
the-t-in-rtf: Fixed.

src/ShareModal.js
the-t-in-rtf: Done.

src/ShareModal.js
the-t-in-rtf: Update to use the right selector (class name).

src/ShareModal.scss
the-t-in-rtf: Done.

src/ShareModal.scss
the-t-in-rtf: Fixed to use `vw` until we can discuss whether to convert the URL to a text input.

src/ShareModal.scss
the-t-in-rtf: Fixed.

src/ShareModal.scss
the-t-in-rtf: I've used your styles for now.

the-t-in-rtf:
I fixed everything but the URL for now, we should discuss that tomorrow.

the-t-in-rtf:
I updated to use the new `ModalBody` component.

chosww:
Thanks @the-t-in-rtf the changes you made look good. About making primary and secondary buttons into more general class name, we also need to update ModalWithFooter component to use regular html button component, instead of using React Bootstrap button component. There are some additional style we need to add to the primary and secondary buttons after changing them to use regular button, which is to add round border to the buttons. We can bring the primary and secondary button styles to index.scss, and we need to change the selector to be button.ClassName__secondaryButton (here I removed .btn after button, because we are not using React Bootstrap button anymore), we also need to find right className for the style. After we change the className, we need to do some clean up, such as assigning right className in ModalWithFooter component, update some test cases with the new className to find primary and secondary buttons, and update the classNames in themes.scss.

I think it makes sense to update the "copy link" button inside of the modal dialog to use this general primary button style, but the one that opens up the share modal actually have different styles with changes I made in C2LC-485 -- share button colours are not the same as primary/secondary in different themes. Also, current style that's applied to the button to render the share icon looks good, so I think we shouldn't convert the share button to use general button style.

the-t-in-rtf:
A lot of the primary / secondary class logic needed to be moved as well, and we needed click/keydown handling, so I created a new `TextButton` component that is similar to but has a slightly different contract than `IconButton`.  This ended up simplifying the styles and meant that we could actually type the `buttonProperties` used in the modal footer.

As suggested, I converted the "copy link" button and removed its duplicate styles, but left the "share" button alone.

src/ShareModal.js
the-t-in-rtf: We reviewed this with Sepi and went with a read only text input.

src/ShareModal.js
```diff
@@ -31,14 +25,10 @@ class ShareCompleteModal extends React.Component<ShareCompleteModalProps, {}> {
         }
     }
 
-    componentDidMount() {
```
the-t-in-rtf: Very good idea, @sbates-idrc.  I was seeing intermittent errors about us accessing the clipboard before there was any user action, and this should prevent that.

PR #303, opened 2021-11-29
C2LC-435: Create modal for different audio features

chosww:
See [C2LC-435](https://issues.fluidproject.org/browse/C2LC-435) for details.


the-t-in-rtf:
I started by testing the work in Chrome, and verified that:

1. The All Audio toggle works
2. The individual toggles work.
3. If you have unticked one type of sound and turn "all sounds" off and on, your sub-preferences are preserved.
4. Save works.
5. Cancel works.

On point 3, I think this is good behaviour, the "all sounds" toggle is kind of like "mute" button, it doesn't destroy your other sound preferences when you eventually "unmute".

However, I'm a little concerned that when "All Sounds" is off, it's not clear that the sub-preferences are not going to be respected, i.e. it suddenly no longer matters if you choose to turn them off or on.  Other preference tools do things like preserve your sub-preferences, but grey them out and don't let you continue to toggle them when they aren't going to be applied.   Here's an example from Firefox:


20211130 sub preferences

I think a similar pattern would more clearly set expectations.  We should review with Sepi.

the-t-in-rtf:
Comparing this to Figma, it seems like the audio icon needs more padding.

Screenshot 2021-11-30 at 10 26 39

src/SoundOptionsModal.test.js
```diff
@@ -0,0 +1,114 @@
...
+        mockOnSave
+    };
+}
+
+function getAudioToggleSwitches(wrapper) {
+    return wrapper.find(ToggleSwitch);
+}
+
+function getCancelButton(wrapper) {
+    return wrapper.find('.ModalWithFooter__secondaryButton');
```
the-t-in-rtf, line 58: This will need to be updated when C2LC-520 (new share modal) is merged.  I'd suggest adding a TODO in both functions.

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,138 @@
...
+            soundOptions.push(
+                <div className='SoundOptions__option-container' key={soundOptionProps.id}>
+                    <div aria-hidden={true}>
+                        {soundOptionProps.label}
+                    </div>
+                    <div className='SoundOptions__toggle-container'>
+                        <div aria-hidden={true}>
+                            {this.props.intl.formatMessage({ id: 'SoundOptionsModal.toggleOff' })}
+                        </div>
+                        <ToggleSwitch
```
the-t-in-rtf, line 88: I was trying to understand this in relation to other recent modals.  Some of them (like actions simplification) prevent the default on events they respond to.  The `ToggleSwitch` used here prevents default for keyboard events, but not clicks.  We should talk about why we prevent default and lightly standardise some of the common components like `ToggleSwitch` and `IconButton` as needed.

src/SoundOptionsModal.test.js
```diff
@@ -0,0 +1,114 @@
...
+        expect(audioToggleSwitches.get(2).props.value).toBe(false);
+    })
+});
+
+describe('When interacting with footer buttons', () => {
+    test('Pressing cancel button should close the modal without changing toggle states', () => {
+        expect.assertions(2);
+        const { wrapper, mockOnCancel } = createMountSoundOptionsModal();
+        const cancelButton = getCancelButton(wrapper);
+        cancelButton.at(0).simulate('click');
```
the-t-in-rtf, line 97: For this test to match the description, you'd need to:

1. Change something
2. Confirm that the change isn't persisted.

You could also just rename it to reflect what it tests, i..e that the cancel button is correctly wired up.

src/SoundOptionsModal.test.js
```diff
@@ -0,0 +1,114 @@
...
+        expect(mockOnCancel.mock.calls.length).toBe(1);
+        expect(mockOnCancel.mock.calls[0][0]).toBe(undefined);
+    });
+    test('Pressing save button should update the toggle states', () => {
```
the-t-in-rtf, line 101: Again, this doesn't seem like what you're actually testing.  You're confirming that the save button is wired up, but you're not actually making a change or confirming that it's passed along on a save.  You can either change the description or (my preference) actually test what you describe.

I used the same pattern in the keyboard input modal work, just testing the functions, but in later work (on the actions simplification menu), it seemed better to actually test the change (or lack of change) as well.

src/App.scss
```diff
@@ -36,25 +36,21 @@ body {
...
+    justify-content: center;
+    column-gap: 0.5rem;
+    padding: 0.5rem 0;
+}
 
-    &:focus {
-        box-shadow: 0 0 0 $c2lc-focus-indicator-width black;
-        outline: none;
+.IconButton.App__header-soundOptions {
+    svg path {
```
the-t-in-rtf, line 52: You could collapse the selector as there's nothing else being updated.

src/Themes.scss
```diff
@@ -1382,6 +1388,17 @@ body.gray-theme {
         }
     }
 
+    .ToggleSwitch:not(.ToggleSwitch--checked).SoundOptionsToggle {
+        background-color: #B9BEC6;
```
the-t-in-rtf, line 1392: Just so I understand, the hover styles are meant to be the same for the gray theme, but not some of the other themes?

src/messages.json
```diff
@@ -208,6 +208,14 @@
         "Scene.outOfBounds": "Scene, in {world}, {numColumns} by {numRows} grid with {character} outside of the scene {relativeDirection} the scene, facing {direction}",
         "Sketchpad.character": "the robot",
         "Sketchpad.label": "A blank sketchbook with grid lines.  Your character in this scene is a Robot.",
+        "SoundOptionsModal.allSounds": "All Sounds",
+        "SoundOptionsModal.musicalSounds": "Musical Sounds",
```
the-t-in-rtf, line 212: We're not entirely consistent about it, but we use alphabetical ordering for a lot of the message keys.  These are almost alphabetical except for this one. 

Alphabetical obviously makes sense to me, but I can also see the sense in putting them it in order by when they're used in the component.  Tthat would match what we do with CSS.

Either way I think it should be organised a little better.

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,138 @@
...
+        return soundOptions;
+    }
+
+    render() {
+        return (
+            <ModalWithFooter
+                show={this.props.show}
+                focusOnOpenSelector={'#sound-options-allsounds'}
+                focusOnCloseSelector={'.App__header-soundOptions'}
+                ariaLabel={this.props.intl.formatMessage({ id: 'SoundOptionsModal.title' })}
```
the-t-in-rtf, line 113: I generally assign the formatted string to a variable if I have to use the same key more than once in the same file.

src/messages.json
chosww: I think it makes sense to just alphabetize everything in messages.json, as the usage order doesn't matter in the file. I updated SoundOptionsModal to be alphabetical order for now.

src/SoundOptionsModal.js
chosww: I changed ariaLabel to aria-labelledBy, as we have clear target for aria-labelledBy.

src/Themes.scss
chosww: Could you tell me more about what other themes you are thinking about?

src/Themes.scss
```diff
@@ -755,6 +755,12 @@ body.gray-theme {
         color: #818A98;
     }
 
+    .IconButton.App__header-soundOptions {
+        svg path {
```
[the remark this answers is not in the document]
chosww: I realized that keyboard modal icon's fill is not #1E1E1E, so I updated the selector to be specific to IconButtons in App__header container.

src/App.scss
chosww: I took out the rule specified to the keyboard modal icon button to be styled by general IconButton styles specified in App__header-menu.

src/SoundOptionsModal.test.js
chosww: Good suggestions, I updated both cancel and save button to check its state changes.

src/SoundOptionsModal.test.js
chosww: I think TODO on these functions are not needed as we will notice test case failure as soon as C2LC-520 gets merged in, but I added them.

chosww:
Thanks @the-t-in-rtf, I talked to Sepideh about the suggestion you made, in regards to the disabling other toggle buttons when All sounds option is toggled off, and she thinks that's a great idea. I also updated the PR based on your feedback.

src/Themes.scss
the-t-in-rtf: I think it's just high contrast that had me confused, as it has different hover styles for checked/not checked.   They all seem fine, it was more of a question.

src/SoundOptionsModal.test.js
the-t-in-rtf: Thanks for humouring me.

the-t-in-rtf: I find TODOs helpful even if I'm just immediately deleting them on the merge.

the-t-in-rtf:
The changes look good.  It's possible to get into a state where audio is technically on, but both types of sound are disabled, which bugged me a little.  On reflection,  I think this is fine, as it errs on the side of preserving the more specific settings when "all sounds" is toggled.

I do think it's important to follow up and persist these settings, I'd propose updating [C2LC-533](https://issues.fluidproject.org/browse/C2LC-533?focusedCommentId=43648&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-43648) to cover that work.

sbates-idrc:
In high contrast theme, focus indication is not visible on the controls in the settings area (sound, theme, keyboard).

I was looking at the designs in Figma but I didn't see a specific design that shows the settings area in high contrast, so this is probably good to ask Sepi about.

I'm thinking maybe that in high contrast the settings area should have a black background, rather than white? That would make it consistent with the panels on the right hand side: in all other themes, the settings area and the panels on the right have the same background, except for high contrast, where the panels on the right have a black background and the settings area has a white background.

sbates-idrc:
`AudioFeedbackToggleSwitch` is no longer used and can be removed:

- `AudioFeedbackToggleSwitch.js`
- `AudioFeedbackToggleSwitch.scss`
- `AudioFeedbackToggleSwitch.test.js`
- Remove the `"AudioFeedbackToggleSwitch.audioFeedback"` message from `messages.json`
- Remove the `AudioFeedbackToggleSwitch` rules from `Themes.scss`

src/App.js
```diff
@@ -908,10 +921,8 @@ export class App extends React.Component<AppProps, AppState> {
         this.focusTrapManager.handleKeyDown(e);
     }
 
-    handleToggleAudioFeedback = (announcementsEnabled: boolean) => {
-        this.setState({
-            announcementsEnabled: announcementsEnabled
-        });
+    handleChangeSoundOptions = (audioEnabled: boolean, announcementsEnabled: boolean, sonificationEnabled: boolean) => {
+        this.setState({ audioEnabled, announcementsEnabled, sonificationEnabled, showSoundOptionsModal: false });
```
sbates-idrc, line 925: We aren't super consistent in our ordering of methods in `App`, but could you move the `handleChangeSoundOptions` method up to be right below `handleSoundOptionsModalClose`, so that the sound options modal handlers are grouped together. Thanks.

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,145 @@
...
+    intl: IntlShape,
+    audioEnabled: boolean,
+    announcementsEnabled: boolean,
+    sonificationEnabled: boolean,
+    show: boolean,
+    onCancel: () => void,
+    onChangeSoundOptions: ( audioEnabled: boolean, announcementsEnabled: boolean, sonificationEnabled: boolean ) => void
+};
+
+type SoundOptionsModalStates = {
```
sbates-idrc, line 24: Please name this type `SoundOptionsModalState` (singular, without the 's' at the end)

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,145 @@
...
+        this.state = {
+            audioEnabled: this.props.audioEnabled,
+            announcementsEnabled: this.props.announcementsEnabled,
+            sonificationEnabled: this.props.sonificationEnabled
+        };
+    };
+
+    handleToggleAudioEnabled = () => {
+        this.setState({
+            audioEnabled: !this.state.audioEnabled
```
sbates-idrc, line 42: Use the value from the callback, rather than inverting the state. To ensure that the toggle switch and state in the component are consistent.

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,145 @@
...
+        });
+    };
+
+    handleToggleAnnouncementsEnabled = () => {
+        this.setState({
+            announcementsEnabled: !this.state.announcementsEnabled
```
sbates-idrc, line 48: Use the value from the callback, rather than inverting the state. To ensure that the toggle switch and state in the component are consistent.

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,145 @@
...
+        });
+    };
+
+    handleToggleSonificationEnabled = () => {
+        this.setState({
+            sonificationEnabled: !this.state.sonificationEnabled
```
sbates-idrc, line 54: Use the value from the callback, rather than inverting the state. To ensure that the toggle switch and state in the component are consistent.

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,145 @@
...
+            announcementsEnabled: this.props.announcementsEnabled,
+            sonificationEnabled: this.props.sonificationEnabled
+        });
+        this.props.onCancel();
+    }
+
+    renderSoundOptions = () => {
+        const soundOptions = [];
+        const soundOptionsProps = [
+            { id:'sound-options-allsounds', label: this.props.intl.formatMessage({ id: 'SoundOptionsModal.allSounds'}), toggleState: this.state.audioEnabled, onChange: this.handleToggleAudioEnabled },
```
sbates-idrc, line 74: Please split these object definitions over multiple lines, with one line per property. To reduce the line lengths and make them easier to read.

src/SoundOptionsModal.test.js
```diff
@@ -0,0 +1,154 @@
...
+        const audioToggleSwitches = getAudioToggleSwitches(wrapper);
+        expect(audioToggleSwitches.length).toBe(3);
+        expect(audioToggleSwitches.get(0).props.id).toBe('sound-options-allsounds');
+        expect(audioToggleSwitches.get(0).props.value).toBe(true);
+        expect(audioToggleSwitches.get(1).props.id).toBe('sound-options-musicalSounds');
+        expect(audioToggleSwitches.get(1).props.value).toBe(true);
+        expect(audioToggleSwitches.get(2).props.id).toBe('sound-options-announcements');
+        expect(audioToggleSwitches.get(2).props.value).toBe(true);
+    });
+    test('Changing audioEnabled, announcementsEnabled and sonificationEnabled properties to false should turn off corresponding toggle siwtches', () => {
```
sbates-idrc, line 84: I don't think that "Changing..." is quite the right description for this test. More like: "When the audioEnabled, announcementsEnabled and sonificationEnabled properties are false, the corresponding toggle switches should be off".

src/SoundOptionsModal.test.js
```diff
@@ -0,0 +1,154 @@
...
+        expect.assertions(8);
+        const { wrapper } = createMountSoundOptionsModal({ audioEnabled: false, announcementsEnabled: false, sonificationEnabled: false });
+        // Other options than allSounds should have disabled class when audioEnabled is false
```
sbates-idrc, line 87: The assertion below is not verifying what this comment states. Only that there are 2 instances of `.SoundOptions__option--disabled`. You don't check where they are.

src/SoundOptionsModal.test.js
```diff
@@ -0,0 +1,154 @@
...
+        expect(getWrapperState(wrapper).sonificationEnabled).toBe(false);
+
+        const allSoundsToggle = getAudioToggleSwitches(wrapper).at(0);
+        allSoundsToggle.simulate('click');
+        expect(getWrapperState(wrapper).audioEnabled).toBe(false);
+
+        const cancelButton = getCancelButton(wrapper);
+        cancelButton.at(0).simulate('click');
+        expect(mockOnCancel.mock.calls.length).toBe(1);
+        // Reset all toggle state to the initial state from props
```
sbates-idrc, line 122: I'll leave a comment also on `SoundOptionsModal.js`, but this isn't quite what we want. The `SoundOptionsModal` attempts to keep its state consistent with that provided in the props by setting them at cancel. However, it is possible to change the announcement setting using a keyboard shortcut, and when this is done, the dialog will show the wrong value. To reproduce:

1. Load Weavly
2. Turn on keyboard shortcuts
2. Verify that announcements are enabled by opening the Delete All dialog
3. Use the keyboard shortcut to disable the announcements (Ctrl + Alt + x, x)
4. Verify that announcements are disabled by opening the Delete All dialog
5. Open the Sound Options dialog

Expected:

* The announcements toggle switch to be off

Actual:

* The announcements toggle switch is on

The fix is to do like in the `WorldSelector` and set the state of the Sound Options dialog when it is shown: https://github.com/codelearncreate/c2lc-coding-environment/blob/818ac96dbffc1b98e08ec0dd8a9b23a82402beb9/src/WorldSelector.js#L138-L144

src/SoundOptionsModal.js
```diff
@@ -0,0 +1,145 @@
...
+    handleOnSave = () => {
+        this.props.onChangeSoundOptions(this.state.audioEnabled, this.state.announcementsEnabled, this.state.sonificationEnabled );
+    }
+
+    handleOnCancel = () => {
+        this.setState({
+            audioEnabled: this.props.audioEnabled,
+            announcementsEnabled: this.props.announcementsEnabled,
+            sonificationEnabled: this.props.sonificationEnabled
+        });
```
sbates-idrc, line 67: Rather than setting the state to the props values at cancel, we need to set them at show. See my comment on `SoundOptionsModal.test.js` which goes into more detail and illustrates a bug with this approach.

src/SoundOptionsModal.test.js
sbates-idrc: And if we want to change toggling of the announcements while the dialog is open, then we would need to check for that also.

src/ToggleSwitch.js
```diff
@@ -38,12 +39,13 @@ export default class ToggleSwitch extends React.Component<ToggleSwitchProps, {}>
...
         return (
             <div
+                aria-disabled={this.props.disabled}
                 id={this.props.id}
                 className={classes}
                 role='switch'
                 aria-label={this.props.ariaLabel}
                 aria-checked={this.props.value}
-                tabIndex='0'
+                tabIndex={this.props.disabled ? -1 : 0}
```
sbates-idrc, line 48: As discussed on Matrix, revert this change and leave disabled ToggleSwitches in the tab order with `tabIndex='0'`

src/SoundOptionsModal.test.js
chosww: ToggleSwitch for "All Sounds" never becomes disabled, and that's why there are 2 instances of .SoundOptions__option--disabled. The comment is saying that when "All Sounds" is toggled off, other two switches will have disabled class, so there are 2 instances. Should I update the test case to be more specific about which toggle become disabled?

src/ToggleSwitch.js
sbates-idrc: Let's put any changes regarding tab order on hold until we resolve what to do in general about disabled controls.

src/SoundOptionsModal.test.js
sbates-idrc: Yes, please. Keep the existing assertion about the number of instances of the `.SoundOptions__option--disabled` class. But then also add checks that they are in the expected places.

PR #304, opened 2021-12-07
C2LC-541: Update world picker to display more than four worlds

chosww:
See [C2LC-541](https://issues.fluidproject.org/browse/C2LC-541) for details.

### DIFF
--- a/src/WorldSelector.js
+++ b/src/WorldSelector.js
@@ -81,41 +81,58 @@ class WorldSelector extends React.Component<WorldSelectorProps, WorldSelectorSta
     }
 
     renderWorldOptions = () => {
-        const worldOptions = [];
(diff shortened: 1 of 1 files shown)


the-t-in-rtf:
I started with QA on this.  Still not very happy with the radio button navigation, but it does work as discussed on chats, down (or right) arrows move down the left column, and then down the right.

I had to zoom way in to make the container scrollable, and checked the scrolling when using keyboard navigation.  Only the radio button and its label are scrolled into view.  We should talk it through and see if we can improve this so that the thumbnail for the highlighted option is also visible.

the-t-in-rtf:
The code changes seem reasonable.  I do wish we could test the column splitting logic a bit, but that kind of thing would be glaringly obvious in QA, so it may be fine not to worry about that.

PR #310, opened 2021-12-10
C2LC-558: Ensure program steps get focus when clicked in Safari

the-t-in-rtf:
See [C2LC-558](https://issues.fluidproject.org/browse/C2LC-558) for details.


chosww:
Thanks @the-t-in-rtf, setting focusCommandBlockIndex to ensure that program steps get focused when clicked makes sense, and I can't think of any side effect of doing this in other browsers either.

PR #314, opened 2021-12-15
C2LC-567: Update character SVG in the scene for Landmarks world and its background

chosww:
See [C2LC-567](https://issues.fluidproject.org/browse/C2LC-567) for details.


the-t-in-rtf:
The changes to the scene character look good.  I wonder if we should also update the world thumbnail to match, might be good to chat w/ Sepi on that.

chosww:
Going with a different approach in C2LC-620.

PR #317, opened 2021-12-21
C2LC-575: Update react-scripts to version 5

sbates-idrc:
See: https://issues.fluidproject.org/browse/C2LC-575


the-t-in-rtf:
We should discuss in today's meeting, I think we can keep the flow coverage reporting if we simply switch to using `flow-coverage-report` via `npx`.  In testing with your branch locally, it looks like the coverage reporting still works that way.    That seems the least effort and we keep the value of the report.

the-t-in-rtf:
I see that the CI checks already used `npx`.  Was there an example of a build that failed or output that was incorrect?  In local testing with node@14, all function as they did before.

sbates-idrc:
Thanks Tony.

This is the failure that I saw in CI:

https://github.com/sbates-idrc/c2lc-coding-environment/runs/4597242223?check_suite_focus=true

And the issue filed against `flow-coverage-report` showing a similar failure:

https://github.com/rpl/flow-coverage-report/issues/208

the-t-in-rtf:
I compared the output of `npm audit` when run against `develop-1.2` and this branch.  This pull resolves 9 moderate severity issues, 10 high, and one critical, leaving us with 2 low severity and 8 medium.

It looks like five more medium severity vulnerabilities would be resolved if we updated node-sass, I filed [C2LC-578](https://issues.fluidproject.org/browse/C2LC-578) to cover that work.  We still have some vulnerabilities brought in by react-scripts, these will be whitelisted when we work on [https://issues.fluidproject.org/browse/C2LC-553](C2LC-553).

the-t-in-rtf:
I compared the packed content sizes in the deployed [https://develop-1-2--c2lc-build.netlify.app](develop-1.2) preview to those in this build, there were no red flags I could find.  The updated version ends up producing a few kilobytes more content, but overall the speed seems fine.

package.json
```diff
@@ -7,18 +7,17 @@
         "classnames": "2.3.1",
         "enzyme": "3.11.0",
         "enzyme-adapter-react-16": "1.15.6",
-        "node-sass": "4.14.1",
+        "node-sass": "7.0.0",
```
the-t-in-rtf, line 10: I didn't see that you had updated this already when I commented earlier and filed [C2LC-578](https://issues.fluidproject.org/browse/C2LC-578).  If we're already updating, then we should go ahead and update to `7.0.1`.

src/CharacterState.js
```diff
@@ -60,6 +60,7 @@ export default class CharacterState {
         return true;
     }
 
+    // eslint-disable-next-line no-use-before-define
```
the-t-in-rtf, line 63: Are there any circumstances in which this check is helpful to us?  If not, maybe we should disable the rule rather than adding all these comments.

the-t-in-rtf:
I left a couple of minor comments, otherwise this looks good.

src/CharacterState.js
sbates-idrc: That's a good question. Disabling the rule, would certainly make the code cleaner by not requiring there comments.

I think the rule offers some value but I think not a huge amount. I'll disable the rule and remove the comments. We can always reconsider. But I think minimizing the number of ESLint-specific comments is worthwhile.

package.json
sbates-idrc: Done. Thanks.

src/CharacterState.js
sbates-idrc: I've disabled the "no-use-before-define" rule

PR #325, opened 2022-01-24
C2LC-581: Loops

sbates-idrc:
Use this PR to track and discuss the overall loops work.

See https://issues.fluidproject.org/browse/C2LC-581


sbates-idrc:
Some notes on terminology. We don't need to act on this now, but I wanted to record some thoughts as they occurred to me.

Current:

* Actions Panel: the name in the user interface for the panel on the left side, containing actions that can be added
* Action: an individual action, such as move forward, or turn left
* Block: we sometimes use block to talk about the physical presentation of an action, such as "the forward 1 block"

With the loops work, we are changing the Program structure. And the code currently looks like:

```
export type ProgramBlock = {
    block: string,
    iterations?: number,
    label?: string
};

export type Program = Array<ProgramBlock>;
```

`ProgramBlock` seems like a reasonable name for the things within a program: the action plus applicable parameters. But I'm thinking that maybe we change the property called `block` currently in `ProgramBlock` to `action`. Though see the next point for a complication.

For loops we have a complication regarding actions and blocks: In the Actions Panel and on the simplification UI, we will have a single action for a loop. But when a loop is part of a program, there are 2 blocks.

This results in the following (at time of writing):

* The Actions Panel, contains an action called `"loop"`, which is serialized by the `AllowedActionsSerializer` as `'l'`
* When a loop is part of a program, there are 2 blocks: `"startLoop"` and `"endLoop"`, which are serialized by the `ProgramSerializer` as `'s'` and `'e'`

So, if `"loop"` is an action (in the Action Panel and simplification UI), are `"startLoop"` and `"endLoop"` also "actions". Or are they something else?

sbates-idrc:
A proposal for discussion:

* Action: the things in the Actions Panel ("loop" is an action)
* Block: the things that programs are made of, in many cases there is a 1-1 correspondence between actions and blocks, but not always (our `ProgramBlock` type)
* Block name or block type: what a block does, either an action name or a block name for actions with multiple blocks (rename the `ProgramBlock` `block` property to `blockName` or `blockType`; "startLoop" and "endLoop" are block names but not actions)

sbates-idrc:
In our design notes (https://docs.google.com/document/d/1afXPFKfmF-lIeNOKIP88neHk3GYN_yx68FzgNB0Mw2c/edit#heading=h.954jagh4excp), we said that we would put the `loopLabel` for `endLoop` blocks in the block `cache`. Currently it is stored directly on the block object.

PR #326, opened 2022-01-26
C2LC-582: Add loop execution to the interpreter

chosww:
See [C2LC-582](https://issues.fluidproject.org/browse/C2LC-582) for details.


src/Interpreter.js
```diff
@@ -86,15 +86,57 @@ export default class Interpreter {
...
+                            resolve();
+                        });
+                    }
+                } else if (command === 'endLoop') {
+                    const { label } = currentProgramStep;
+                    if( label != null ) {
+                        const currentProgram = programSequence.getProgram();
+                        let programCounter = programSequence.getProgramCounter();
+                        let isNested = false;
+                        for (let i = 0; i < programSequence.getProgramCounter(); i++) {
```
sbates-idrc, line 104: An alternative algorithm:

* Start iterating through the program backwards from the current program counter
* Until we reach the corresponding startLoop
* Then stop
* Then we don't need the `isNested` variable: if we encounter a startLoop that isn't the corresponding one, we reset its `iterationsLeft`, otherwise, we have reached as far as we need to go

src/Interpreter.js
```diff
@@ -86,15 +86,57 @@ export default class Interpreter {
...
+                                        isNested = true;
+                                        programCounter = i;
+                                    // When there's no more iterations left, increment the programCounter
+                                    } else {
+                                        programCounter += 1;
+                                    }
+                                } else {
+                                    // When startLoop has a different label and if it is a nested loop; reset its iterationsLeft
+                                    if (isNested) {
+                                        currentProgram[i].iterationsLeft = currentProgram[i].iterations;
```
sbates-idrc, line 120: We shouldn't modify the program in place like this. Instead make a shallow copy of the program to operate on, using `slice()`, and use `Object.assign()` to make a new block object with the updated `iterationsLeft`.

src/ProgramSequence.js
```diff
@@ -111,4 +122,22 @@ export default class ProgramSequence {
...
         return false;
     }
+
+    decrementLoopIterations(loopLabel: string): ProgramSequence {
+        const program = this.program.slice();
+        for (let i = 0; i < program.length; i++) {
+            const { block, iterations, iterationsLeft, label } = program[i];
+            if (block === 'startLoop' && label === loopLabel) {
+                if (iterationsLeft != null && iterationsLeft > 0) {
+                    program[i] = {
```
sbates-idrc, line 132: As we talked about, change this to use `Object.assign()`

PR #327, opened 2022-02-01
C2LC-594: Converted to tracking 'disallowed' actions.

the-t-in-rtf:
See [C2LC-594](https://issues.fluidproject.org/browse/C2LC-594) for details.


src/App.js
```diff
@@ -943,14 +938,14 @@ export class App extends React.Component<AppProps, AppState> {
     handleToggleAllowedCommand = (event: Event, commandName: CommandName) => {
```
sbates-idrc, line 938: The `handleToggleAllowedCommand()` method isn't used and can be removed.

src/DisallowedActionsSerializer.js
```diff
@@ -6,7 +6,7 @@
 */
 import type { ActionToggleRegister } from './types';
 
-export default class AllowedActionsSerializer {
+export default class DisallowedActionsSerializer {
```
sbates-idrc, line 9: The source code for `DisallowedActionsSerializer` contains a few references to 'allowed' rather than 'disallowed' that need to be updated.

Such as the file level comment:

```
We use the same conventions as the program to store a register of "allowed actions" in the URL, so that the choices
made in the "simplification" menu are persisted.
```

I guess this comment should also be updated as "same conventions as the program" isn't strictly true any longer.

And some variables, such as:

```
let serializedAllowedActions = '';
for (const [actionKey, isAllowed] of Object.entries(actionToggleRegister)) {
    if (isAllowed) {
```

the-t-in-rtf: I fixed these.  I also searched for, found, and fixed a few others.

src/App.js
the-t-in-rtf: Fixed.

src/ActionsMenuItem.test.js
```diff
@@ -112,15 +112,15 @@ it("Allows reenabling 'used' items that are hidden.", () => {
...
-    [  true   , false, false  ],
-    [  true   , true , true   ],
-])("The menu item is disabled when 'allowed' and 'used'.",
-    (allowed: boolean, used: boolean, expectedDisabled: boolean) => {
+    // disallowed, used , expectedDisabled
+    [  true , false, false  ],
+    [  true , true , false  ],
+    [  false, false, false  ],
+    [  false, true , true   ],
+])("The menu item is disabled when 'disallowed' or 'used'.",
```
sbates-idrc, line 120: This description is not accurate. It was accurate as it was: "The menu item is disabled when 'allowed' and 'used'."

src/ActionsMenuItem.js
```diff
@@ -54,14 +54,14 @@ export class ActionsMenuItem extends React.Component< ActionsMenuItemProps, {} >
             icon = React.createElement(iconType);
         }
 
-        // Disable the checkbox if the button is allowed and used.
-        const isCheckboxDisabled = (this.props.isAllowed && this.props.isUsed);
+        // Disable the checkbox if the button is disallowed or used.
```
sbates-idrc, line 57: This comment is not accurate. It was accurate as it was: "Disable the checkbox if the button is allowed and used."

PR #329, opened 2022-02-02
C2LC-593

michelled:
Updating the keyboard shortcuts docs.


the-t-in-rtf:
I went through the images first, here are my observations:

> Turn the audio feedback announcements on or off

Right now, this is "Not available".  You should be able to take a screenshot of the announcement toggle for this (see notes below).

> Add the selected action block to the program after the focused node or the focused action block

This would be better if one or the other of the mentioned elements had focus in the image.  I'd suggest taking a shot of the same area with the focus on the add node, as that's more commonly what people will see.  The "hover" style might also be a clearer choice here.

> Delete the focused action block from the program

I'm less certain on this one, but it seems like you could use a screenshot of the action panel (with the trash can focused) for a single program block, as in:

Screenshot 2022-02-03 at 13 13 19

Another option would be to use the "hover" style for something you're activating, as the user would see using the mouse:

Screenshot 2022-02-03 at 13 15 42

> Move focus to the add node toggle

This is the only focus one you have a screenshot for, and I'm not sure whether we should have more of them, or none.  With the current screenshot, you might also think that the shortcut will toggle that button, which it won't.  If we want to keep this, IMO we should have screenshots for the results of the action, i.e. the element should be focused in the screen shot, as in:

Screenshot 2022-02-03 at 13 12 38

> Move focus to the beginning of the program

Similarly, this one is fairly easy to demonstrate in a small screenshot.

Screenshot 2022-02-03 at 13 07 14

In general, I'm not sure we should have shots for all the focus shortcuts.  If we feel that communicates the intent, displaying the results seems reasonable.  If we have images for the "focus" shortcuts, they need to be clearly distinguishable.  The best example is "Move focus to the beginning of the program" versus "Add the selected action block to the beginning of the program".

One option would be to use the "hover/click" styles for all elements that are being activated ("Add"), and the "focus" styles for those that are receiving focus ("Move focus").

We should talk it through with the team.

the-t-in-rtf:
> Turn the character left 45 degrees

You already do exactly what I was talking about above with this screenshot, which I find really clear.  Hover styles FTW.

src/messages.json
```diff
@@ -147,7 +147,7 @@
         "KeyboardInputModal.Description.increaseProgramSpeed": "Press {key} to make the program play faster.",
         "KeyboardInputModal.Description.playPauseProgram": "Press {key} to play or pause the program.",
         "KeyboardInputModal.Description.refreshScene": "Press {key} to refresh the scene.",
-        "KeyboardInputModal.Description.showHide": "Press {key} to show / hide the keyboard shortcuts menu.",
+        "KeyboardInputModal.Description.showHide": "Press {key} to show the keyboard shortcuts menu.",
```
the-t-in-rtf, line 150: Fair point, that used to be the case, I think?

the-t-in-rtf: Either way, good catch.

src/GenerateKeyboardDoc.test.js
```diff
@@ -0,0 +1,128 @@
+// Generates the keyboard Markdown document
+//
+// Note: this file assumes that the working directory is the project directory.
+//
+// To generate the keyboard Markdown document file:
+//
+// - Set 'generateDoc' to true below
+// - Run this test with: npm test GenerateKeyboardDoc
```
the-t-in-rtf, line 8: Unless we plan to do this as part of our tests in CI, this doesn't need to be a test.

It seems better to convert this to a node script.  You'd then make a separate npm script definition for this, i.e. `npm generateDoc`.

Once it's no longer a test, you could get rid of the `generateDoc` variable.

src/KeyboardInputSchemes.js
```diff
@@ -108,43 +108,58 @@ const ExtendedKeyboardSequences: KeyboardInputScheme = {
...
-                actionName: "focusAppHeader"
+                actionName: "focusAppHeader",
+                description: "Move focus to the page title"
             },
             addNodeToggle: {
                 keyDef: { code: "KeyT", key: "t" },
-                actionName: "focusAddNodeToggle"
+                actionName: "focusAddNodeToggle",
+                description: "Move focus to the add node toggle",
+                icon: "A23.png",
```
the-t-in-rtf, line 123: I would suggest naming all of these a little more specifically.  For example, we'll want to remove the left180 and right180 definitions once those are no longer focusable in the UI.  It should be at least vaguely clear to someone doing that or similar work what each image represents.  Just a little bit of context is fine, like `focus-add-node-toggle.png` here.

src/GenerateKeyboardDoc.test.js
the-t-in-rtf: I can see that this is more involved than I'd like because node is not great with imports.  One option would be to move the definitions to a json(5) file, and import that into `KeyboardInputSchemes.js`.

the-t-in-rtf: (and into your doc generation script).

michelled: I wonder if I should pull this out as a separate issue. In the current code base, we generate the docs by uncommenting code inside the test - this seemed like a good interim solution.

src/messages.json
michelled: Thanks @the-t-in-rtf. I've switched to the icons that Sepideh gave me based on the conversation we had. I've also changed the filenames to be more understandable.

src/GenerateKeyboardDoc.test.js
the-t-in-rtf: Yeah, this is fine for now, given how much work we'd have to do to do it any other way.

the-t-in-rtf:
A few of the "move" focus screenshots don't have any focus styles:

1. "Move focus to the add node toggle"
2. "Move focus to the pen toggle"
3. "Move focus to the world selector"

The first two in particular are important to distinguish, as we will eventually add shortcuts to activate each of those controls.

All of the "select the ... block" screenshots lack hover styles, but I don't think that matters, as we're not distinguishing between focus and click (hover) for those.

Other than that, the updates are an improvement and look good.

PR #331, opened 2022-02-08
C2LC-584: Hide replace button for start and end loop blocks.

the-t-in-rtf:
See [C2LC-584](https://issues.fluidproject.org/browse/C2LC-584) for details.

### DIFF
--- a/src/ActionPanel.js
+++ b/src/ActionPanel.js
@@ -101,6 +101,9 @@ class ActionPanel extends React.Component<ActionPanelProps, {}> {
 
     render() {
         const stepInfoMessage = this.makeStepInfoMessage();
+        const currentStepName = this.props.programSequence.getProgramStepAt(this.props.pressedStepIndex).block;
+        const isLoopBlock = ["startLoop", "endLoop"].indexOf(currentStepName) !== -1;
+
         return (
             <React.Fragment>
                 <div className="ActionPanel__background">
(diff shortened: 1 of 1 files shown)


the-t-in-rtf:
This should be reviewed after C2LC-583 and C2LC-594.  I'm waiting to take care of the ARIA labels until 583 is closer to being merged (or merged).

sbates-idrc:
I have started a `C2LC-584` branch with some updates:

https://github.com/sbates-idrc/c2lc-coding-environment/commits/C2LC-584

* Merged `develop-1.2` into `C2LC-584`
* Updated the logic for not showing the replace button for loop labels
  * Rather than use a `hidden` property, I changed the code to not render the button at all
  * Hiding the button, but keeping it in the document, was causing the tab navigation to get stuck

The `ActionPanel` container is still rendered at the same width with 3 buttons as 4, which looks a little odd. But I think at this stage we could tackle that separately. I'm happy to merge this as is.

Please have a look at my changes and let me know if they look good.

sbates-idrc:
I just checked the design and in the design, the control panel stays the same width but the button is just missing. We can do that also. We can check with Sepi what she wants. But I think we can still merge this as is so we can focus on getting 1.2 ready for the co-design sessions next week.

sbates-idrc:
I guess another option is just to disable and leave the button there, like we do for move left and right. We should talk with Sepi.

PR #335, opened 2022-02-14
C2LC-585: Update startLoop and endLoop blocks

chosww:
See [C2LC-585](https://issues.fluidproject.org/browse/C2LC-585) for details.

### DIFF
--- a/src/CommandBlock.js
+++ b/src/CommandBlock.js
@@ -20,11 +20,11 @@ import { ReactComponent as Loop } from './svg/Loop.svg';
 
 type CommandBlockProps = {
     commandName: string,
-    onClick: (evt: SyntheticEvent<HTMLButtonElement>) => void,
     disabled: boolean,
     looplabel?: string,
     loopiterationsleft?: ?number,
-    className?: string
+    className?: string,
+    onClick: (evt: SyntheticEvent<HTMLButtonElement>) => void,
 };
 
 // TODO: Revise this once there is a proper strategy for typing SVG-backed
--- a/src/CommandBlock.js
+++ b/src/CommandBlock.js
@@ -17,13 +17,13 @@ import { ReactComponent as Right45 } from './svg/Right45.svg';
 import { ReactComponent as Right90 } from './svg/Right90.svg';
 import { ReactComponent as Right180 } from './svg/Right180.svg';
 import { ReactComponent as Loop } from './svg/Loop.svg';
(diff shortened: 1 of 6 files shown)


src/CommandBlock.js
```diff
@@ -17,14 +17,14 @@ import { ReactComponent as Right45 } from './svg/Right45.svg';
...
-import { ReactComponent as LoopStart } from './svg/LoopStart.svg';
-import { ReactComponent as LoopEnd } from './svg/LoopEnd.svg';
 
 type CommandBlockProps = {
     commandName: string,
-    onClick: (evt: SyntheticEvent<HTMLButtonElement>) => void,
     disabled: boolean,
-    className?: string
+    looplabel?: string,
+    loopiterationsleft?: ?number,
```
sbates-idrc, line 25: Please update the names of the new props to use camelCase: `loopLabel` and `loopIterationsLeft`

src/CommandBlock.js
```diff
@@ -57,15 +55,32 @@ export default React.forwardRef<CommandBlockProps, Button>(
...
         const iconType = commandBlockIconTypes.get(commandName);
         if (iconType) {
-            icon = React.createElement(
+            children = React.createElement(
                 iconType,
                 {
                     className: 'command-block-svg'
                 }
             );
+        } else {
```
sbates-idrc, line 67: For our future selves, I think it would be better here to check explicitly that `commandName` is `startLoop` or `endLoop` to guard this rendering of the loop blocks, rather than assuming that we have a loop if the command isn't found in `commandBlockIconTypes`.

sbates-idrc:
I'm seeing something a little funky with timing of change of styles in high contrast theme. I'll try and describe how to produce.

In high contrast theme, if you move your mouse pointer on and off of a start loop block, there is a slight timing difference between changing the colour of the whole block background and the background of the label. This causes the circular label background, which isn't normally visible in high contrast, to be visible for a brief moment.

For high contrast, could we make the circular label background transparent maybe?

sbates-idrc:
The difference in timing is present for the other themes also, it's just not as noticeable as for high contrast because in high contrast it makes the circular label background briefly visible when otherwise it isn't.

sbates-idrc:
I added a note to our QA doc regarding the not handling of the "MM" label. We can delay addressing that and not block merging of this PR.

PR #341, opened 2022-02-22
C2LC-599: Move loop decrement to endLoop

sbates-idrc:


chosww:
Thanks @sbates-idrc, the changes and updated test cases look good. With the new iterationsLeft decrement logic made expects in the test case to just check endLoop reduced number of expects a lot 👍

PR #342, opened 2022-02-23
C2LC-601: Support program step replacement with loop command

chosww:
See [C2LC-601](https://issues.fluidproject.org/browse/C2LC-601) for details.

### DIFF
--- a/src/ProgramSequence.js
+++ b/src/ProgramSequence.js
@@ -156,7 +156,25 @@ export default class ProgramSequence {
 
     overwriteStep(index: number, command: string): ProgramSequence {
         const program = this.program.slice();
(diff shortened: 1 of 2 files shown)


src/ProgramSequence.js
```diff
@@ -156,7 +156,25 @@ export default class ProgramSequence {
 
     overwriteStep(index: number, command: string): ProgramSequence {
```
the-t-in-rtf, line 157: Is there any reason not to type this as `CommandName` now that we have it?

the-t-in-rtf: We discussed in chats, we don't currently have a set of values for what's used here (and in the program block value).

PR #354, opened 2022-03-09
C2LC-614: Add keyboard shortcut to select loop command

chosww:
See [C2LC-614](https://issues.fluidproject.org/browse/C2LC-614) for details.


chosww:
@the-t-in-rtf @sbates-idrc, I used 's' key for selecting loop command, and the shortcut is grouped under "selectActionChanges" which has all other selection shortcuts. I know that 's' key doesn't really make sense, but 'l' key is used for turn left commands. I have a few different ideas that maybe better than using 's' key for loop, and they are:

1. Update existing action selection shortcuts for movements to use 'm' key, rather than 'a' key, and update loop command to use 'c' + 'l', as loop command is under controls section of the action panel. 
2. Update existing action selection shortcuts for left and right to be under 'a' + 't' combination, as they are turning commands, and we already use combination 'c' + 't' for character position turn shortcuts. 

Both approaches cause existing on screen keyboard to be updated, as they shortcuts are getting updated, so for now, I just opened a PR with 's' key, but I would like to hear about what you think.

sbates-idrc:
As discussed at our meeting on March 15, we will use the following:

* Ctrl+Alt+x, m, f
* Ctrl+Alt+x, m, b
* Ctrl+Alt+x, m, l, 1
* Ctrl+Alt+x, m, l, 2
* Ctrl+Alt+x, m, r, 1
* Ctrl+Alt+x, m, l, 2
* Ctrl+Alt+x, c, l

Additionally, as Ctrl+Alt+x, c is currently used by the character positioning shortcuts, we will change those to use Ctrl+Alt+x, p,...

src/App.js
```diff
@@ -796,6 +796,12 @@ export class App extends React.Component<AppProps, AppState> {
                         case("selectRight90"):
                             this.setState({ "selectedAction": "right90" });
                             break;
+                        case("selectRight180"):
+                            this.setState({ "selectedAction": "right180" });
+                            break;
```
sbates-idrc, line 801: We have removed turn right 180 action and don't need the "selectRight180" shortcut anymore.

src/KeyboardInputSchemes.js
```diff
@@ -36,6 +36,8 @@ export type ActionName =
     | "selectLeft90"
     | "selectRight45"
     | "selectRight90"
+    | "selectRight180"
```
sbates-idrc, line 39: We have removed turn right 180 action and don't need the "selectRight180" shortcut anymore.

PR #357, opened 2022-03-10
C2LC-611: Handle URLs that are too long

sbates-idrc:


chosww:
Thanks @sbates-idrc for the simplification and refactoring work. As we talked about in our general chat, I did some more manual testings to verify the work, and I couldn't find any issues. I tried old commands (move 2 and 3 steps) too, and your implementation handles them well.

PR #363, opened 2022-03-25
C2LC-626: Pause for stepTimeMs on empty loop iterations

sbates-idrc:


the-t-in-rtf:
Just a note for the future, the fix for this ended up decrementing the loop counter even for empty loops, so the expected behaviour is for an empty loop with 99 iterations to pause 99 times.

PR #370, opened 2022-03-31
C2LC-630: Remove the ability to unselect actions

chosww:
See [C2LC-630](https://issues.fluidproject.org/browse/C2LC-630) for details.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -627,12 +627,10 @@ export class App extends React.Component<AppProps, AppState> {
         });
     };
 
-    handleCommandFromCommandPalette = (command: ?string) => {
-        if (command) {
-            this.setState({
-                selectedAction: command
-            });
-        }
+    handleCommandFromCommandPalette = (command: string) => {
+        this.setState({
+            selectedAction: command
+        });
     };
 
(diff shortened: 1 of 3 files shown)


sbates-idrc:
I think that it would be worthwhile to go a little further with this change. The `CommandPaletteCommand` still has the toggle behaviour. I think we can remove it from there also.

We can rename the `onChange` callback prop of `CommandPaletteCommand` and make the parameter non-null:

`onSelect: (commandName: string) => void`

And simplify the `CommandPaletteCommand` click handler to:

```
handleClick = () => {
    this.props.onSelect(this.props.commandName);
};
```

As we don't need it to toggle any more.

Then we can also remove the truthy check from `App.handleCommandFromCommandPalette`:

```
handleCommandFromCommandPalette = (command: string) => {
    this.setState({
        selectedAction: command
    });
};
```

sbates-idrc:
And update the toggle test in `CommandPaletteCommand.test.js` to verify that clicking does not toggle.

chosww:
Thanks @sbates-idrc, I cleaned up CommandPaletteCommand's toggle behaviour.

PR #371, opened 2022-04-04
C2LC-631: Update starting position of the characters when they are moved by CharacterPositionController

chosww:
See [C2LC-631](https://issues.fluidproject.org/browse/C2LC-631) for details.

### DIFF
--- a/src/Scene.js
+++ b/src/Scene.js
@@ -17,6 +17,8 @@ export type SceneProps = {
     characterState: CharacterState,
     theme: ThemeName,
     world: WorldName,
+    startingX: number,
+    startingY: number,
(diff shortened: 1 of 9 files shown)


src/Scene.js
```diff
@@ -365,6 +367,24 @@ class Scene extends React.Component<SceneProps, {}> {
                             {grid}
                             <g clipPath='url(#Scene-clippath)'>
                                 {this.drawCharacterPath()}
+                                <rect
+                                    // Dotted rectangle to indicate starting grid cell
```
sbates-idrc, line 371: Maybe something more general like: "Cell outline to indicate starting grid cell"

PR #372, opened 2022-04-05
C2LC-591: Reset loop label to A when there are no loops in the program

chosww:
See [C2LC-591](https://issues.fluidproject.org/browse/C2LC-591) for details.

### DIFF
--- a/src/ProgramSequence.js
+++ b/src/ProgramSequence.js
@@ -45,6 +45,16 @@ export default class ProgramSequence {
         return this.program[index];
     }
 
+    hasLoopBlock(): boolean {
+        for (const programBlock of this.program) {
+            const { block } = programBlock;
+            if (block === 'startLoop') {
+                return true;
+            }
+        }
+        return false;
+    }
+
     currentStepIsControlBlock(): boolean {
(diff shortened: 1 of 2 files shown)


src/ProgramSequence.js
```diff
@@ -45,6 +45,16 @@ export default class ProgramSequence {
         return this.program[index];
     }
 
+    hasLoopBlock(): boolean {
+        for (const programBlock of this.program) {
+            const { block } = programBlock;
+            if (block === 'startLoop') {
```
sbates-idrc, line 51: I think rather than using destructuring here, we can just use:

    if (programBlock.block === 'startLoop') {

src/ProgramSequence.test.js
```diff
@@ -14,6 +14,29 @@ test('ProgramSequence constructor', () => {
     expect(programSequence.getLoopIterationsLeft()).toBe(loopIterationsLeft);
 });
 
+describe('Test hasLoopBlock', () => {
+    test('When the program has at leas one loop block', () => {
```
sbates-idrc, line 18: Spelling: 'at least'

PR #373, opened 2022-04-11
C2LC-632: Add keyboard shortcuts to move focus by program blocks

chosww:
See [C2LC-632](https://issues.fluidproject.org/browse/C2LC-632) for details.


sbates-idrc:
Please move the new code to focus the previous and next program block to their own methods, to reduce the amount of code in the keyboard shortcut case statement. So that the case statement looks something like:

```
...
case("focusPreviousProgramBlock"):
    this.focusPreviousProgramBlock();
    break;
case("focusNextProgramBlock"):
    this.focusNextProgramBlock();
    break;
...
```

PR #374, opened 2022-04-19
C2LC-640: Remember character direction and update starting position indicator colour in DeepOcean world

chosww:
See [C2LC-640](https://issues.fluidproject.org/browse/C2LC-640) for details.


src/App.js
```diff
@@ -1055,11 +1058,11 @@ export class App extends React.Component<AppProps, AppState> {
         return commandBlocks;
     }
 
-    makeStartingCharacterState(world: WorldName, startingX: number, startingY: number): CharacterState {
+    makeStartingCharacterState(world: WorldName, startingDirection: number, startingX: number, startingY: number): CharacterState {
```
sbates-idrc, line 1061: We can remove the `world` parameter from `makeStartingCharacterState` as we are no longer using it. Previously we used it to retrieve the direction, but we are now passing the direction in explicitly.

src/Utils.test.js
```diff
@@ -41,24 +41,24 @@ test('Test getWorldFromString', () => {
 });
 
 test('Test getStartingPositionFromString', () => {
-    // (1, 1)
-    expect(getStartingPositionFromString('aa', 16, 8, 2, 3)).toStrictEqual({ x: 1, y: 1 });
+    // (1, 1) facing east
+    expect(getStartingPositionFromString('aab', 16, 8, 2, 3, 2)).toStrictEqual({ x: 1, y: 1, direction: 2 });
```
sbates-idrc, line 45: For these tests, please use a `defaultDirection` that is different from any of the other numbers being tested, to make sure that we are using the right parameter. Such as 7 (rather than 2).

sbates-idrc:
Please update the handlers for the "up", "right", "down", and "left" character positioning controls to also set the `startingDirection` state (like for the "turnLeft" and "turnRight" handlers).

PR #383, opened 2022-07-20
C2LC-654: Starting to add the Haunted House world.

michelled:


michelled:
This pull request is not ready - I'm making it so that folks can see the work in progress.

michelled:
> This pull request is not ready - I'm making it so that folks can see the work in progress.

The pull request is now ready for review!

src/Worlds.scss
```diff
@@ -95,6 +95,113 @@ body.contrast-theme {
     }
 }
 
+/* Haunted House */
```
sbates-idrc, line 98: `Worlds.scss` is a little out of alphabetical order (Landmarks is in the wrong place). For this PR, could you please move the new styles for Haunted House down to between Deep Ocean and Jungle please? That way just Landmarks is in the wrong place and can be moved later.

michelled: I moved the styles for both Haunted House and Landmarks so the file is now alphabetical by world name.

PR #387, opened 2022-07-28
C2LC-657: Updated the characters for Marble Run and Sketchpad.

michelled:


sbates-idrc:
It's fine. We can leave the naming as is. Thanks.

PR #390, opened 2022-08-29
C2LC-668: Add logic to select voice for announcement

sbates-idrc:
See https://issues.fluidproject.org/browse/C2LC-668

### DIFF
--- a/src/AudioManagerImpl.js
+++ b/src/AudioManagerImpl.js
@@ -205,6 +205,7 @@ export default class AudioManagerImpl implements AudioManager {
         right90: Sampler,
         right180: Sampler
     };
+    speechSynthesisVoices: Array<SpeechSynthesisVoice>;
 
     constructor(audioEnabled: boolean, announcementsEnabled: boolean, sonificationEnabled: boolean) {
         this.audioEnabled = audioEnabled;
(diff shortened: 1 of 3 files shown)


sbates-idrc:
I added some timing instrumentation around the call to `selectSpeechSynthesisVoice()` to see if retrieving the list of voices for every announcement was expensive. I found that it was not expensive. Details below.

```
> git diff -U4
diff --git a/src/AudioManagerImpl.js b/src/AudioManagerImpl.js
index 9400c1a..fad6335 100644
--- a/src/AudioManagerImpl.js
+++ b/src/AudioManagerImpl.js
@@ -232,15 +232,19 @@ export default class AudioManagerImpl implements AudioManager {
             const messageId = "Announcement." + messageIdSuffix;
             const toAnnounce = intl.formatMessage({ id: messageId}, messagePayload);
             const utterance = new SpeechSynthesisUtterance(toAnnounce);

+            const timeBeforeMs = Date.now();
             // TODO: When we support non-English UI language(s),
             //       ensure that the language is specified correctly
             utterance.voice = selectSpeechSynthesisVoice(
                 'en',
                 window.navigator.language,
                 window.speechSynthesis.getVoices()
             );
+            const timeAfterMs = Date.now();
+            /* eslint-disable no-console */
+            console.log(`Time to select voice: ${timeAfterMs - timeBeforeMs} ms`);

             window.speechSynthesis.speak(utterance);
         }
     }
```

I ran the above code on the following devices and browsers:

* Windows Chrome
* Windows Firefox
* Windows Edge
* Mac Safari
* Mac Chrome
* Mac Firefox
* iPad Safari
* Chromebook

For every case, the reported time was 0 ms or 1 ms, except for Mac Safari and iPad Safari where the first call took longer (Mac 470 ms, iPad 277 ms) but then subsequent calls each took 0 ms or 1 ms.

PR #391, opened 2022-08-30
C2LC-671: Update character position from the position coordinate boxes only when there's intentional changes

chosww:
See [C2LC-671](https://issues.fluidproject.org/browse/C2LC-671) for details.


sbates-idrc:
We no longer set the starting position when the character is moved. Setting the starting position is now a separate operation. Closing.

PR #393, opened 2022-10-04
C2LC-672: Add Matomo analytics to Weavly

sbates-idrc:


chosww:
Thanks @sbates-idrc, the code changes look good and I didn't find any issues while doing some manual testings; however, I wonder if the notification needs some additional properties to make it accessible.

PR #395, opened 2022-11-08
C2LC-674: Update hover styles for ThemeSelector options in different themes

chosww:
See [C2LC-674](https://issues.fluidproject.org/browse/C2LC-674) for details.


sbates-idrc:
For high contrast on the simplification dialog, when a row has both hover and focus, we show some white padding between the focus outline and the hover background.

image

sbates-idrc:
On grayscale, the focus outline is not black, but a dark grey.

PR #443, opened 2024-02-06
Only recognise painting motions that start within the scene (fixes #438)

### LINKED ISSUE #438: Painting motions must start within the scene

sbates-idrc:


sbates-idrc:
We have decided not to implement this behaviour, and instead implement tile selection on mouse down, in addition to click. See https://github.com/codelearncreate/c2lc-coding-environment/pull/445

PR #535, opened 2024-07-24
Replace node-sass with Dart Sass

sbates-idrc:


greatislander:
[This deprecation](https://sass-lang.com/documentation/breaking-changes/mixed-decls/) was only introduced in Sass 1.77.7 so pinning the Sass version at 1.77.6 resolves the error. For the `abs` deprecation there is no way around it but I think it can safely be silenced based on how the Sass `abs` function is being used.

greatislander:
Re: silencing deprecations: https://sass-lang.com/documentation/js-api/interfaces/options/#silenceDeprecations

sbates-idrc:
I have updated dependencies since making this PR and I have created a new PR with those updated dependencies plus Dart Sass 1.77.6: https://github.com/codelearncreate/c2lc-coding-environment/pull/544

PR #541, opened 2024-09-26
Add ProgramBlockCache and new ProgramBlock types

sbates-idrc:
- Replace the Map used for the program block cache with a new `ProgramBlockCache` class, with typed accessors for the loop label and position

- Add new separate program block types for: `MovementProgramBlock`, `StartLoopProgramBlock`, and `EndLoopProgramBlock`


chosww (approved):
Thanks @sbates-idrc, the changes look good.

